PR #1, opened 2015-05-20
Options and `basePath` addition

gmac:
Good work, this plugin is definitely what I needed. A few additions though:

* Make manifest filename configurable through constructor options.
* Adding `basePath` option for prefixing all filepaths within the manifest. This is useful for including the build output location within all file references.


danethurber:
this is something i meant to add be didn't get around to it. thanks for the help

PR #4, opened 2015-07-01
Added transformExtensions option for sourcemaps

jordansexton:
ManifestPlugin currently doesn't handle sourcemap assets properly. `main.js.map` and `main.css.map` will result in a single output key, `main.map`. This is caused by the expectation of only a single file extension.

This is probably not the best way to solve this, but it works with minimal code changes. The `imageExtensions` RegExp was also changed to handle this. Additional extensions can easily be added for other modifiers. The change also includes a passing test for

Note: this branch is based off of [patch-1](https://github.com/jordansexton/webpack-manifest-plugin/tree/patch-1) (#3). I can change that and rebase if you like the PR but don't want to have to cherry pick around.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -9,7 +9,7 @@ function ManifestPlugin(opts) {
     stripSrc: null,
     imageExtensions: /(jpe?g|png|gif|svg)$/i
   }, opts || {});
-};
+}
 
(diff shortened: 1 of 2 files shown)


danethurber:
awesome. thanks for the help

PR #6, opened 2015-07-28
Emit manifest with assets

tikotzky:
This fixes an issue I came across when trying to use this plugin with [`webpack-dev-middleware`](https://github.com/webpack/webpack-dev-middleware)

`webpack-dev-middleware` uses [`memory-fs`](https://github.com/webpack/memory-fs) instead of writing to disk. 
The issue was that when the plugin tried to write the manifest [here](https://github.com/danethurber/webpack-manifest-plugin/blob/b29088eaf942500a19587f826f37cc34ded5f33e/lib/plugin.js#L73) it would crash because the output dir would not exist since webpack was writing the files to memory.

I changed the plugin so it emits the `manifest.json` as a standard asset so that webpack will write it using whichever fs it is currently using.

regarding a test...
I started writing a specific test for running with `memory-fs` but then realized that if I change all tests to use `memory-fs` I can completely remove the need `rimraf`. 
This way compatibly with `memory-fs` is being tested without an explicit test.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -1,5 +1,3 @@
-var fs = require('fs');
-var path = require('path');
 var _ = require('lodash');
 
 function ManifestPlugin(opts) {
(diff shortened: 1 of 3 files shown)


neilgupta:
Thank you! This is exactly the fix I was looking for to get webpack-manifest-plugin working with karma-webpack.

tikotzky:
Thanks!

danethurber:
thank you

PR #8, opened 2015-10-01
Don't include the basePath in manifest keys

bensmithett:
We usually want to reference an asset by some key we know (its filename, e.g. `foo.js`) & get its real public path (e.g. `https://some.cdn.url/foo.1234.js`) in return.

If the basePath gets added to both the key & value in the manifest, it's impossible to do this.

Since it's a simple change I thought I'd just PR it rather than opening an issue, but curious if you have a use case for including basePath in the keys?

:smile: cheers!


armanm:
What's supposed to happen if two assets with the same name exist on two different paths?

bensmithett:
OK, I did something daft & don't actually need this. Never mind!

PR #13, opened 2015-11-26
require('fs'); closes #12

jordansexton:

PR #14, opened 2015-11-27
Handling of non-image, non-chunk assets; readFile changed

jordansexton:
This removes the sketchy extension whitelist for image types (refs. #11), and uses webpack's compilation events to solve a longstanding problem, how to map output assets possibly including hashes to inputs. I've run the tests and refactored until everything passed.

Edit: ~~This includes the two commits from #13, but I can cut them if there's a merge conflict.~~
It also overrides #11, but it actually solves the problem at the same time (manifest assets get uniquely associated with user inputs).

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,3 +46,4 @@ new ManifestPlugin({
 * `fileName`: The manifest filename in your output directory (`manifest.json` by default).
 * `basePath`: A path prefix for all file references. Useful for including your output path in the manifest.
 * `stripSrc`: removes unwanted strings from source filenames
+* `cache`: In [multi-compiler mode](https://github.com/webpack/webpack/tree/master/examples/multi-compiler) webpack will overwrite the manifest on each compilation. Passing a shared `{}` as the `cache` option into each compilation's ManifestPlugin will combine the manifest between compilations.
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -1,3 +1,4 @@
+var path = require('path');
 var _ = require('lodash');
 
 function ManifestPlugin(opts) {
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -36,7 +36,7 @@ ManifestPlugin.prototype.apply = function(compiler) {
 
   compiler.plugin('emit', function(compilation, compileCallback){
     var stats = compilation.getStats().toJson();
-    var assetsByChunkName = stats.assetsByChunkName;
+
     var manifest;
     if (this.opts.readFile && fs.existsSync(outputPath)) {
       manifest = JSON.parse(fs.readFileSync(outputPath).toString());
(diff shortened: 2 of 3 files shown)


jordansexton:
I added a test for the new `cache` option and documented it. This is a breaking change from the `readFile` option but there's no alternative, as the file can't be read after #6 -- it isn't written until all compilations are emitted.

Sorry for the messy PR and issues. As I was digging into the original fix I discovered the other problems and refactored a lot.

shamrin:
@jordansexton Nice patch! However, I've found two problems when I ran your code against our project (I've tested my PR #11 with the same project):

The first one is serious. The seconds is minor and probably not even a bug…

#### 1. Crash

The code from this PR fails with it encounters nameless chunks:

```
/Users/user/src/webpack-manifest-plugin/lib/plugin.js:47
      var chunkName = chunk.name.replace(this.opts.stripSrc, '');
                                ^

TypeError: Cannot read property 'replace' of null 
```

I've changed the line above to:

```js
      var chunkName = (chunk.name || '').replace(this.opts.stripSrc, '');
```

and it works, but then it generates strange key for this nameless chunk:

```
{
  ".js": "1.1.02cfefee0f5fbac7d2af.js",
…
}
```

#### 2. Manifest keys don't use the names set with file-loader plugin.

Extract from our webpack config:

```js
            {
                test: /\.png$/,
                loader: 'file-loader',
                query: {
                    name: '[1]-[name].[hash:20].png',
                    regExp: 'node_modules/([^/]+)/'
                }
            }
```

This PR generates the following manifest:

```
{
…
  "layers-2x.png": "leaflet-layers-2x.324b4fcaf164735c6272.png",
  "layers.png": "leaflet-layers.7cb0d2482ecadc1b80eb.png",
  "marker-icon-2x.png": "leaflet-marker-icon-2x.1c824216f354218b04b2.png",
  "marker-icon.png": "leaflet-marker-icon.87f6ca46ac356e81dc43.png",
  "marker-shadow.png": "leaflet-marker-shadow.e7bd5e4b8dbbc3dfbe3e.png",
…
}
```

With the master branch I get the following manifest:

```
{
…
  "leaflet-layers-2x.png": "leaflet-layers-2x.324b4fcaf164735c6272.png",
  "leaflet-layers.png": "leaflet-layers.7cb0d2482ecadc1b80eb.png",
  "leaflet-marker-icon-2x.png": "leaflet-marker-icon-2x.1c824216f354218b04b2.png",
  "leaflet-marker-icon.png": "leaflet-marker-icon.87f6ca46ac356e81dc43.png",
  "leaflet-marker-shadow.png": "leaflet-marker-shadow.e7bd5e4b8dbbc3dfbe3e.png"
…
}
```

Notice how the first version misses `leaflet-` prefix. `Leaflet` refers to http://leafletjs.com. You can probably see this behaviour yourself if you do `npm install leaflet` and `require` in your app.

But I don't know what behaviour is correct…

D1plo1d:
+1 for this PR. With this PR my manifest.json properly includes webpack's font assets (eg. `fontawesome-webfont.eot?v=4.5.0`).

mastilver:
Hi @jordansexton 

I'm trying to add tests arround: 

```js
     moduleAssets[file] = path.join(
          path.dirname(file),
          path.basename(module.userRequest)
      );
```

But all the tests are wrote are passing with both the code above and this one:

```js
moduleAssets[file] = file;
```

Can you give me a scenario when you code is needed?

PR #15, opened 2016-01-05
Limit image asset name to base name to allow path in name

jmurzy:
This allows one to set a different output path for images (vs. what's defined for output.path).

"main.js": "bundles/main.js",
"assets/images/xyz.svg": "assets/images/xyz.svg"

should be

"main.js": "bundles/main.js",
"xyz.svg": "assets/images/xyz.svg"

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -1,4 +1,5 @@
 var _ = require('lodash');
+var path = require('path');
 
 function ManifestPlugin(opts) {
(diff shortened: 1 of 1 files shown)


danethurber:
Try again with the latest release and let me know if you're still having issues.

jmurzy:
@danethurber I'm having the same issue with the latest release.

Issue #16, opened 2016-02-18
Nameless chunks result in an exception


rgrove:
As of commit 2ff2950f (which landed in version 1.0.0), webpack-manifest-plugin throws the following exception when it encounters a nameless chunk:

```
/path/redacted/node_modules/webpack-manifest-plugin/lib/plugin.js:43
      var chunkName = chunk.name.replace(this.opts.stripSrc, '');
                                ^

TypeError: Cannot read property 'replace' of null
    at ManifestPlugin.<anonymous> (/path/redacted/node_modules/webpack-manifest-plugin/lib/plugin.js:43:33)
    at Array.reduce (native)
    at ManifestPlugin.<anonymous> (/path/redacted/node_modules/webpack-manifest-plugin/lib/plugin.js:42:39)
    at Compiler.applyPluginsAsync (/path/redacted/node_modules/tapable/lib/Tapable.js:71:13)
    at Compiler.emitAssets (/path/redacted/node_modules/webpack/lib/Compiler.js:226:7)
    at Compiler.<anonymous> (/path/redacted/node_modules/webpack/lib/Compiler.js:184:10)
    at /path/redacted/node_modules/webpack/lib/Compiler.js:403:12
    at Compiler.next (/path/redacted/node_modules/tapable/lib/Tapable.js:67:11)
    at Compiler.<anonymous> (/path/redacted/node_modules/webpack/lib/CachePlugin.js:40:4)
    at Compiler.applyPluginsAsync (/path/redacted/node_modules/tapable/lib/Tapable.js:71:13)
```

It looks like @shamrin called this out as a bug in PR #14, but for some reason it was never fixed and the PR was merged anyway.

stevenp:
+1 on this. To avoid this error I had to change:
```
require.ensure([], (require) => {
  cb(null, require('../containers/FormContainer/FormContainer.js'));
});
```
to:
```
require.ensure([], (require) => {
  cb(null, require('../FormContainer/FormContainer.js'));
}, 'Form');
```
@rgrove: On a side note, I had a heck of a time figuring out that this plugin was the source of my error. How did you get that stack trace? I was running webpack through Grunt, and only got the `TypeError` with no additional detail.

rgrove:
@stevenp I'm running the webpack CLI directly here. It's been a while since I last used Grunt, but I seem to recall it hides stack traces unless you pass `--stack` on the command line. Maybe that's why you didn't see a stack trace?

[...]

PR #17, opened 2016-04-12
Nameless chunks don't throw errors

gmac:
Just updated to the latest version of Webpack Manifest Plugin and noticed what looks to be a widely reported bug. Nameless chunks cause errors... see: https://github.com/danethurber/webpack-manifest-plugin/issues/16.

Nameless chunks are actually fairly common in Webpack, because they're created by the native code splitting feature `require.ensure(...)`. Code splits will produce new chunks formatted as:

```
{
  name: null,
  files: [ '4.4.9cfa2b27bb561f8c2bbd.js' ]
}
```

Now, these nameless code split chunks are automatically catalogued within the Webpack runtime so that it knows how to fetch and reconstitute the chunk composition. These nameless units are not entry chunks nor initial chunks, so there's not much value in being able to load them independently via a manifest. Therefore, this update simply omits nameless chunks from the manifest and leaves them as wards of the Webpack runtime.


jordansexton:
While I like the philosophical approach of treating nameless chunks as a hidden implementation detail, there are reasons it would be valuable to list them as keys in the manifest still. For example, I use the manifest on the server side to look up files for serving in memory (avoiding a check for the file's existence, which is relevant when using Android's virtualized filesystem for bundled assets). I'm of the belief that the manifest should report all files output by Webpack, even if they aren't entries or initial chunks. What about listing them by chunk ID, e.g.

`"4.4": "4.4.9cfa2b27bb561f8c2bbd"`

or just filename, if we don't want to infer or care about the chunk ID:

`"4.4.9cfa2b27bb561f8c2bbd": "4.4.9cfa2b27bb561f8c2bbd"`

I realize this is duplicative and not very useful if one is just looking for entry chunks, but I think it's preferable for the manifest to be complete than opinionated in that way. What do you think?

danethurber:
sorry i've been busy and just getting time to address this. 

@gmac thanks for a quick solution

@jordansexton are you suggesting something along these lines to get the manifest key if there is no chunk.name

```
var chunkName = chunk.name
  ? chunk.name.replace(this.opts.stripSrc, '')
   : chunk.id.toString();
```

proposal -> https://github.com/danethurber/webpack-manifest-plugin/commit/cfb3fd80ab88162927b8aebdcfb53fa307536518

gmac:
For something like that, it would probably be more reliable just to look
through the chunk files. Nameless chunks should still have all
generated file information there, at which time a literal mapping of the
webpack file name to itself could be established.

gmac:
Okay @jordansexton, makes sense to me. Having the manifest document all files –known and unknown– is a good idea. For known files, manifest gives us a mapping of what we called a file relative to what Webpack called it. For unknown files, manifest gives us a record of all files we never knew about that Webpack created and named for us. I've updated the PR to make nameless chunks output a literal mapping of their child files, such as:

 ```
{ "1.1.js": "1.1.js" }
```

Also adds a new test.

jordansexton:
Nice! Thanks for the thoughtful fix to the issue I (rather carelessly) introduced. 😅

PR #18, opened 2016-04-14
Add subfolder support

orbiteleven:
I export my files into subfolders:

```
...
output: {
    filename: 'javascripts/main-[hash].js',
  },
...
```

however this (great!) plugin was not picking up on that for `main.js` (and `main.css`), though it would for assets:

```
...
"images/foo.png": "images/foo-79a318e1a4e1bc0aac94b6ba2bbb3b2a.png",
"main.js": "javascripts/main-2ce5bbf6dc7dbe06a0e9.js",
"main.css": "stylesheets/main-2ce5bbf6dc7dbe06a0e9.css"
...
```

This diff will prepend any subfolders to `main.js`/`main.css`:

```
...
"images/foo.png": "images/foo-79a318e1a4e1bc0aac94b6ba2bbb3b2a.png",
"javascripts/main.js": "javascripts/main-2ce5bbf6dc7dbe06a0e9.js",
"javascripts/main.css": "stylesheets/main-2ce5bbf6dc7dbe06a0e9.css"
...
```

Not as familiar with the webpack plugin system, so if there's a "smarter" way to do this, by all means let me know :)

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -11,6 +11,13 @@ function ManifestPlugin(opts) {
   }, opts || {});
 }
 
+ManifestPlugin.prototype.getFileName = function(chunkName, file) {
+  var dir = path.dirname(file);
+  var type = this.getFileType(file);
+  var filename = chunkName + '.' + type;
+  return path.join(dir, filename);
+}
+
 ManifestPlugin.prototype.getFileType = function(str) {
   str = str.replace(/\?.*/, '');
   var split = str.split('.');
@@ -43,7 +50,7 @@ ManifestPlugin.prototype.apply = function(compiler) {
       var chunkName = chunk.name.replace(this.opts.stripSrc, '');
 
       return chunk.files.reduce(function(memo, file){
-        memo[chunkName + '.' + this.getFileType(file)] = file;
+        memo[this.getFileName(chunkName, file)] = file;
         return memo;
       }.bind(this), memo);
     }.bind(this), {}));
--- a/spec/plugin.spec.js
+++ b/spec/plugin.spec.js
@@ -143,6 +143,20 @@ describe('ManifestPlugin', function() {
         done();
       });
     });
+
+    it('respects subfolders', function(done) {
+      webpackCompile({
+        entry: path.join(__dirname, './fixtures/file.js'),
+        output: {
+          filename: 'javascripts/main.js'
+        }
+      }, function(manifest){
+        expect(manifest).toBeDefined();
+        expect(manifest['javascripts/main.js']).toBeDefined();
+        expect(manifest['javascripts/main.js']).toEqual('javascripts/main.js');
+        done();
+      });
+    })
   });
 
   describe('with ExtractTextPlugin', function(){


pgilad:
Great addition, but this should be added with a flag to be backwards compat.

pgilad:
@danethurber is it possible to advance this somehow?

mblarsen:
@danethurber cool, could you add the flag/option so it can me merged?

lukeed:
Doesn't look like there's much activity here, but this should 💯 be included.

spec/plugin.spec.js
```diff
@@ -143,6 +143,20 @@ describe('ManifestPlugin', function() {
...
     });
+
+    it('respects subfolders', function(done) {
+      webpackCompile({
+        entry: path.join(__dirname, './fixtures/file.js'),
+        output: {
+          filename: 'javascripts/main.js'
+        }
+      }, function(manifest){
+        expect(manifest).toBeDefined();
```
mastilver: can you follow `expect(manifest).toEqual(...)`

mastilver:
I agree this PR make sense, though I think it's a breaking change

Can you rebase that on top of master, please?

lib/plugin.js
```diff
@@ -43,7 +50,7 @@ ManifestPlugin.prototype.apply = function(compiler) {
       var chunkName = chunk.name.replace(this.opts.stripSrc, '');
 
       return chunk.files.reduce(function(memo, file){
-        memo[chunkName + '.' + this.getFileType(file)] = file;
+        memo[this.getFileName(chunkName, file)] = file;
```
mastilver: Wouldn't`memo[file] = file` be enough here?

mastilver:
This is already possible by doing:

```js
{
        entry: {
          'javascripts/main': path.join(__dirname, './fixtures/file.js')
        },
        output: {
          filename: '[name].js'
        }
      }

```

lukeed:
You've missed something 😄 The entry filename is supposed to be a file _name_ only. It affects the rest of the build cycle & will likely break other plugins.

mastilver:
@lukeed Cool 👍 

Now is #66 enough for that or should we add another options like https://github.com/danethurber/webpack-manifest-plugin/issues/23#issuecomment-275038728 suggested ?

I'm not a big fan on adding yet another option, while #66 will allow:

```
map: (file) => {
  return {
    ...file,
    name: path.dirname(file.path) + file.name
  }
}
```

lukeed:
Not sure tbh. I've been tuned out of this for quite some time. All this time I've been using [my fork](https://github.com/lukeed/webpack-manifest-plugin) and it's done exactly what I need -- and what's described in this PR.

orbiteleven:
Yeah... This is an over-year-old PR. I've moved on to other solutions and not really following the work on this project. Is this still needed?

lukeed:
Last I checked was 3 to 4 months ago. Was certainly still needed then. I'm in same boat -- my own workaround has been great.

orbiteleven:
Original diff was 14 April 2016; renewed interest in March, but I'm not sure if #66 or #23 solve this in different ways. Again, not using the project anymore. I'm happy to update the diff, but it'd take some "dusting off" on my part

mastilver:
@orbiteleven I added your test on #66: https://github.com/danethurber/webpack-manifest-plugin/pull/66/files#diff-4170f24227285478db88bb820df9a581R446

mastilver:
My issue with adding this PR in is that everybody got their own needs and we can't take care of all of them. This is why the hooks provided by #66 are to me the solution for that

mastilver:
Thank you @orbiteleven but I'm closing this as it's now possible using the map option:
https://github.com/danethurber/webpack-manifest-plugin/blob/master/spec/plugin.spec.js#L545-L566

orbiteleven:
Works for me!

PR #19, opened 2016-04-15
Resolving relative paths

mendlik:
Hi. I added a mechanism that resolves relative paths.

Scenario:
My opts.basePath path is '/assets' but I needed to put favicon in the root '../'.
Webpack-manifest-plugin did not resolve relative paths so I added it.


mblarsen:
@mendlik nice PR. Seems to be a conflict.

mendlik:
Sorry, I've already walked around it.

PR #28, opened 2016-09-26
Option to emit to file system

andreicek:
When using this plugin is conjunction with [webpack-dev-server](https://webpack.github.io/docs/webpack-dev-server.html) it will emit to memory only. This is preferred but in some use cases (eg. integration with rails) you also need the manifest file on the filesystem.

I've added a flag `writeToFileEmit` that will write the JSON output to the build folder as well.


andreicek:
cc/ @DarkoKukovec

PR #31, opened 2016-10-16
Add webpack-dev-server support

martynovs:
feature: skip hot updates from manifest
fix: basePath was added to already cached manifest values on each compilation


michaek:
Great timing, @martynovs - I was just about to look into this and make a PR! These changes look good to me, and have worked in my testing.

PR #38, opened 2017-05-08
Resolve paths

Lobstrosity:
When populating the manifest, resolves `basePath`/`publicPath` with `key`/`value` (to resolve parent directory references).

So `/base/subfolder/../file.ext` becomes `/base/file.ext`.

Side benefit is trailing slashes on `basePath`/`publicPath` are now optional since `join` takes care of that for you.

Note that it uses `path.normalize(path.join())` instead of `path.resolve()` so that ancestors of `basePath`/`publicPath` are not exposed. (The approach could be simplified if that’s not a concern.) This results in invalid paths if an absolute path is specified with several parent directory references. E.g., `resolvePath('/absolute/../..', 'file.ext')` returns `'/file.ext'` (not valid), whereas `resolvePath('relative/../..', 'file.ext')` returns `'../file.ext'`.

<hr />

My particular use case was with `sass-loader` plus `file-loader` to export CSS-referenced images.
* `sass-loader` processing SCSS to `/static/css`.
* `file-loader` processing images with `outputPath=../images/` (which was relative to CSS output, so `/static/images`).
* `webpack-manifest-plugin` with `basePath` of `/static/css/` so that the manifest paths were absolute.

Before this change, an image would appear in the manifest as `/static/css/../images/file.ext`. With this change, it appears as `/static/images/file.ext`.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -14,6 +14,10 @@ function ManifestPlugin(opts) {
   }, opts || {});
 }
 
+function resolvePath() {
+  return path.normalize(path.join.apply(path, arguments));
+}
+
 ManifestPlugin.prototype.getFileType = function(str) {
   str = str.replace(/\?.*/, '');
   var split = str.split('.');
@@ -75,7 +79,7 @@ ManifestPlugin.prototype.apply = function(compiler) {
     // This allows output path to be reflected in the manifest.
     if (this.opts.basePath) {
       manifest = _.reduce(manifest, function(memo, value, key) {
-        memo[this.opts.basePath + key] = this.opts.basePath + value;
+        memo[resolvePath(this.opts.basePath, key)] = resolvePath(this.opts.basePath, value);
         return memo;
       }.bind(this), {});
     } else if (this.opts.publicPath) {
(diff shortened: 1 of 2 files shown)


mastilver:
@Lobstrosity Could you rebase on top on `master`

Tests are failing on my side

mastilver:
@a-x- 

> Tests are failing on my side

Also are you :100: it's not a breaking change?

a-x-:
I not 100% sure, but I saw passed tests (after rebase as I understood) and code likes good
It can breaks usage relied on buggy behavior probably.

Excuse me if you want to test it better

mastilver:
fine :) can you fix the tests please

mastilver:
@a-x- And next time can you squash it instead of merging the PR, it makes thinks look nicer :)

a-x-:
hey
there was a deep evening at here yesterday :)
I see you fix tests yourself now.

I checked locally, everything passes 👍
I just don't understand why was tests in CI good

mastilver:
>  I just don't understand why was tests in CI good

Weird, I think you got confuse with code approval checks :p

mastilver:
This caused an issue see #38 , we need to find a workaround to fix this issue while keeping urls safe

PR #43, opened 2017-06-22
feat(chunks): add chunks option to filter chunks out

mastilver:


mastilver:
This will be doable with #55

PR #45, opened 2017-06-27
Copy webpack plugin integration

davidmpaz:
Hi,

first of all, thanks for this project. I am writing to ask whether could be possible to include this patch in main stream or in case is not good, lets try to find/coordinate a solution for include copied asset in manifest.

My use case is that I have several images used in php templates, these are not compiled/processed by Webpack and as such they are never copied to target directory. 

In this scenario, plugins like [copy-webpack-plugin](https://github.com/kevlened/copy-webpack-plugin) are very helpful. One missing feature, I think many people will appreciate, is to have those copied assets in the manifest file also.

Some notes on my findings:

* variable [moduleAssets](https://github.com/danethurber/webpack-manifest-plugin/blob/v1.1.0/lib/plugin.js#L67) never contain copied assets, making check on `name` variable fail.
* Still we have assets in `compilation.getStats().toJson().assets`.
* In case you ask why to use plugin instead a file-loader. I can say I tried, but the solution I was coming into was not good to me. I needed to mix build logic into code that is run inside application etc..

Many thanks in advance and best regards.


weaverryan:
@davidmpaz could you add a test for this? Does it work in all the situations you've found? Or are there some caveats?

mastilver:
I'm sorry you need to find another solution for that, it will output 'normal' file twice

I'm also not sure about the option, ideally I would like to land this on `v2` turned on by default

mastilver:
@davidmpaz If you test again the latest develop and `includeCopiedAssets` tests will fail

davidmpaz:
Hi @mastilver precisely today I was trying to add some tests and figure it out of that duplication with tests. I should have had added tests since the beginning. Test are failing in travis now but I will fix them soon. Today I can not continue.

@weaverryan I don't see any other caveats besides the issue with duplication mentioned, which should be fixed now.

mastilver:
@davidmpaz No problem at all :)

Take your time, I'm planning to merge your PR for `v2` which is not happening anytime soon
I will review your existing work. Pick it up only when you are ready ;)

lib/plugin.js
```diff
@@ -63,10 +64,15 @@ ManifestPlugin.prototype.apply = function(compiler) {
 
     // module assets don't show up in assetsByChunkName.
     // we're getting them this way;
+    var allowCopied = this.opts.includeCopiedAssets;
```
mastilver: Remove the option, it will added as a breaking change, so no need to be backward compatible

spec/plugin.spec.js
```diff
@@ -221,4 +221,48 @@ describe('ManifestPlugin', function() {
       });
     });
   });
+
+  describe('assets added by third parties to "compilation.assets"', function() {
+    it('are included also when enabled.', function(done) {
+
+      var options = webpackConfig({
```
mastilver: Can you not use `webpackConfig`?
Follow the same pattern as `with ExtractTextPlugin`

spec/plugin.spec.js
```diff
@@ -221,4 +221,48 @@ describe('ManifestPlugin', function() {
...
+        manifestOptions: {
+            includeCopiedAssets: true
+        }
+      });
+
+      // add plugin to get third party assets included
+      let Mock = require(path.join(__dirname, './plugin-mock'));
+      options.plugins.unshift(new Mock());
+
+      webpackCompile(options, function(manifest, stats) {
```
mastilver: `webpackCompile(options, {}, function(manifest, stats) {`

spec/plugin.spec.js
```diff
@@ -221,4 +221,48 @@ describe('ManifestPlugin', function() {
...
+        expect(manifest['third.party.js']).toEqual('third.party.js');
```
mastilver: Can you follow this pattern: `expect(manifest).toEqual(...)`, it will prevents future issues from slipping through

spec/plugin.spec.js
```diff
@@ -221,4 +221,48 @@ describe('ManifestPlugin', function() {
...
+    it('are included also when enabled.', function(done) {
+
+      var options = webpackConfig({
+        manifestOptions: {
+            includeCopiedAssets: true
+        }
+      });
+
+      // add plugin to get third party assets included
+      let Mock = require(path.join(__dirname, './plugin-mock'));
```
mastilver: `let Mock = require('./plugin-mock');` and can yoou put it at the very beginning of the file (Maybe find a better name ;) )

spec/plugin.spec.js
```diff
@@ -221,4 +221,48 @@ describe('ManifestPlugin', function() {
...
+        done();
+      });
+    });
+  });
+
+  describe('assets added by third parties to "compilation.assets"', function() {
+    it('are not included when disabled.', function(done) {
+
+      var options = webpackConfig({
+        entry: {
```
mastilver: use `context: __dirname` and `one: ''./fixtures/file.js''`

spec/plugin.spec.js
```diff
@@ -221,4 +221,48 @@ describe('ManifestPlugin', function() {
...
+      options.plugins.unshift(new Mock());
+
+      webpackCompile(options, function(manifest, stats) {
+        expect(manifest['third.party.js']).toEqual('third.party.js');
+        done();
+      });
+    });
+  });
+
+  describe('assets added by third parties to "compilation.assets"', function() {
```
mastilver: No need for this test, as I want to make it a default

lib/plugin.js
davidmpaz: Ok. understandable.

spec/plugin.spec.js
davidmpaz: Did this since could not find a way to use mock plugin only for one test. Now it is added to all, directly in initial configuration.

spec/plugin.spec.js
davidmpaz: Done :)

spec/plugin.spec.js
davidmpaz: Fixed

spec/plugin.spec.js
davidmpaz: I would like to have a way to test it without using the plugin. Any idea how best to do. Currently I always add it.

spec/plugin.spec.js
davidmpaz: does it make sense to test when no assets are added (or not found) in entry?

lib/plugin.js
```diff
@@ -24,6 +24,11 @@ ManifestPlugin.prototype.getFileType = function(str) {
   return ext;
 };
 
+ManifestPlugin.prototype.existNameInManifest = function(name, manifest) {
```
mastilver: :/ I'm not sure about this one... Why is that needed?

spec/plugin.spec.js
```diff
@@ -85,7 +88,8 @@ describe('ManifestPlugin', function() {
       }, {}, function(manifest) {
         expect(manifest).toBeDefined();
         expect(manifest).toEqual({
-          'main.js': 'main.js'
+          'main.js': 'main.js',
+          'third.party.js': 'third.party.js'
```
mastilver: I'm curious where does `third.party.js` come from?

spec/copy-plugin-mock.js
```diff
@@ -0,0 +1,25 @@
...
+            source: function() {
+                return compiledMock;
+            }
+        };
+
+        callback();
+    });
+};
+
+module.exports = FakeCopyWebpackPlugin;
```
davidmpaz: thanks for the suggestion! I have added a new tool to my belt :) ... Applied the formatting to relevant files.

lib/plugin.js
davidmpaz: It is not enough to test whether we have the entry in manifest keys, values need checking too. 

As far i see, we have assets in [chunk.files](https://github.com/danethurber/webpack-manifest-plugin/blob/v1.1.2/lib/plugin.js#L50) and in [stats.assets](https://github.com/danethurber/webpack-manifest-plugin/blob/v1.1.2/lib/plugin.js#L66) storing processed ones in `moduleAssets` and `manifest`, in `moduleAssets` we receive nothing, which was initial motivation of this PR, then in `manifest` i tested first only keys for avoiding duplication, it results we need more.

When using hashed values in assets name, key testing is not enough, I need to check whether file name is not in value also. For example for tests using hashes and base path, i got manifests like:
```
{ 'one.1067da9bcb5b3a357a83.js': 'one.1067da9bcb5b3a357a83.js',
  'one.js': 'one.1067da9bcb5b3a357a83.js',
  'third.party.js': 'third.party.js' }
```
the first entry in that object is wrong.

mastilver: Oh... I see

spec/plugin.spec.js
davidmpaz: The purpose of this PR is to allow any third party plugin to add assets to manifest using `compilation.assets` from webpack. That is what [copy-webpack-plugin does here](https://github.com/kevlened/copy-webpack-plugin/blob/v4.0.1/src/writeFile.js#L84)

This mean that if we execute such plugins before manifest plugin executes, then manifest plugin will be able to pick those assets and add them to final output. In this case hashing i think must rely on those external plugins since files are not processed by webpack at all.

to answer the real question: `third.party.js` simulates such behaviour, the file is added to compilation* by some plugin (`copy-plugin-mock.js`) not by webpack itself when requiring/importing assets as normal.

I don't know whether i missed some edge case, so please if you think something else must be taken into account, a bit more explanation will be welcome. I have based this on finding through debugging, not through deep knowledge of webpack :)

lib/plugin.js
mastilver: What do you think about doing this instead:

```js
else if (asset.chunks.length === 0){
  memo[asset.name] = asset.name;
}
```

spec/plugin.spec.js
davidmpaz: One of the reason I tried to use first `webpackConfig` was to be able to run the specs with/without adding the mocking plugin. If you have an idea how could this be added (testing with and without using plugin), please give me some hints and I can added the code for it. Right now is always used in specs, i think it would be a nice to have in tests.

lib/plugin.js
davidmpaz: Please could you explain a bit more. I can not follow the idea.

spec/plugin.spec.js
mastilver: Oh I see, I didn't realised you added it to all tests

mastilver: Can you create a new test that follow the same logic as `with ExtractTextPlugin` but for `FakeCopyWebpackPlugin`

spec/copy-plugin-mock.js
```diff
@@ -0,0 +1,26 @@
+function FakeCopyWebpackPlugin() {
```
mastilver: Move this file to `helpers/copy-plugin-mock.js`

lib/plugin.js
```diff
@@ -14,7 +14,7 @@ function ManifestPlugin(opts) {
   }, opts || {});
 }
 
-ManifestPlugin.prototype.getFileType = function(str) {
+ManifestPlugin.prototype.getFileType = function (str) {
```
mastilver: leave it like this `function(str)`

spec/plugin.spec.js
davidmpaz: Ok i will do. I have time now.

spec/copy-plugin-mock.js
davidmpaz: ok.

lib/plugin.js
davidmpaz: uhmm. i think i can not trust on idea IDEs for formatting using editorconfig rules. I will look on editorconfig for a rule to add such restriction. Otherwise other will incur in same error

README.md
```diff
@@ -53,3 +53,4 @@ new ManifestPlugin({
 * `stripSrc`: removes unwanted strings from source filenames (string or regexp)
 * `writeToFileEmit`: If set to `true` will emit to build folder and memory in combination with `webpack-dev-server`
 * `cache`: In [multi-compiler mode](https://github.com/webpack/webpack/tree/master/examples/multi-compiler) webpack will overwrite the manifest on each compilation. Passing a shared `{}` as the `cache` option into each compilation's ManifestPlugin will combine the manifest between compilations.
+* `includeCopiedAssets`: Allow to include assets that where not processed by webpack but included into assets by plugins like [copy-webpack-plugin](https://github.com/kevlened/copy-webpack-plugin)
```
mastilver: remove this

lib/plugin.js
mastilver: Rather than checking if a value is already present in the manifest (Which I can see could conflict with #55 , just make sure that the asset is a real asset, not something generated by webpack (make sure that `asset.chunks` is empty). Not sure if I make sense?

spec/plugin.spec.js
```diff
@@ -6,6 +6,8 @@ var _ = require('lodash');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var extractTextPluginMajorVersion = require('extract-text-webpack-plugin/package.json').version.split('.')[0];
 var plugin = require('../index.js');
+// add plugin to get third party assets included
+var FakeCopyWebpackPlugin = require(path.join(__dirname, './helpers/copy-plugin-mock'));
```
mastilver: `var FakeCopyWebpackPlugin = require('./helpers/copy-plugin-mock');`

lib/plugin.js
davidmpaz: To be honest, this is the point in which my knowledge of webpack start shortening. I am not sure if it makes sense neither.

Something i did notice when debugging was that files added by copy plugin were never coming in `compilation.chunks`, but in `compilation.stats.assets`. Does this answer your question?

Unfortunately I could not extract much knowledge from issue #55 too many use cases i don't know at all. Which part do you think will conflict?

mastilver:
Hi @davidmpaz 

Can you let me know if you have time to carry on with this one, I want to release `2.x` next week-end

Can you remove all your style changes: no space between function and parentheses 

Thank you :)

davidmpaz:
Hi @mastilver,

sure. I did notice that you merged to update branches but thought you were going to continue working on it.

I will take a look to finish it between today and tomorrow, otherwise I will not have time later :)

cheers

mastilver:
Great, thank you@ :)

Let me know if you can't so I can carry on

One last thing, for those files, can you see `isModuleAsset` to `false`, please, that way there is a way to tell them apart

davidmpaz:
Hi @mastilver,

after looking better to current work I decided would be easier to just create another pull request than re-basing this one. The changes have been to much to conciliate again what was there :)

Please lets continue in another PR I will made now.

PR #48, opened 2017-07-03
fix: Only include initial chunks

### LINKED ISSUE #46: Do not add code splitting bundles into manifest
I want to use `require.ensure` for separating some large modules as DLL.
But `webpack-manifest-plugin` add it into manifest 
so chunks are loading immediately

a-x-:
closes #46 _Do not add code splitting bundles into manifest_

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -42,24 +42,26 @@ ManifestPlugin.prototype.apply = function(compiler) {
     var stats = compilation.getStats().toJson();
     var manifest = {};
 
-    _.merge(manifest, compilation.chunks.reduce(function(memo, chunk) {
-      var chunkName = chunk.name ? chunk.name.replace(this.opts.stripSrc, '') : null;
+    _.merge(manifest, compilation.chunks
+      .filter(chunk => chunk.isInitial ? chunk.isInitial() : chunk.initial)
+      .reduce(function(memo, chunk) {
+        var chunkName = chunk.name ? chunk.name.replace(this.opts.stripSrc, '') : null;
 
-      // Map original chunk name to output files.
-      // For nameless chunks, just map the files directly.
-      return chunk.files.reduce(function(memo, file) {
-        // Don't add hot updates to manifest
-        if (file.indexOf('hot-update') >= 0) {
+        // Map original chunk name to output files.
+        // For nameless chunks, just map the files directly.
+        return chunk.files.reduce(function(memo, file) {
+          // Don't add hot updates to manifest
+          if (file.indexOf('hot-update') >= 0) {
+            return memo;
+          }
+          if (chunkName) {
+            memo[chunkName + '.' + this.getFileType(file)] = file;
+          } else {
+            memo[file] = file;
+          }
           return memo;
-        }
-        if (chunkName) {
-          memo[chunkName + '.' + this.getFileType(file)] = file;
-        } else {
-          memo[file] = file;
-        }
-        return memo;
-      }.bind(this), memo);
-    }.bind(this), {}));
+        }.bind(this), memo);
+      }.bind(this), {}));
 
     // module assets don't show up in assetsByChunkName.
     // we're getting them this way;
--- a/spec/plugin.spec.js
+++ b/spec/plugin.spec.js
@@ -216,7 +216,7 @@ describe('ManifestPlugin', function() {
         output: { filename: '[name].[hash].js' }
       }, function(manifest, stats) {
         expect(manifest['nameless.js']).toEqual('nameless.'+ stats.hash +'.js');
-        expect(manifest['1.1.'+ stats.hash +'.js']).toEqual('1.1.'+ stats.hash +'.js');
+        expect(manifest['1.1.'+ stats.hash +'.js']).toEqual(undefined);
         done();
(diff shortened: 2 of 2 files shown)


mastilver:
Do we need the `omitDlls` options?

Couldn't you use `chunk.initial` or `chunk.isInitial()`?

a-x-:
What is chunk.initial or chunk.isInitial()?
I cannot find them in the source code or documentation

mastilver:
is a boolean on chunk that tell if they are required on initial run

`.filter(chunk => chunk.isInitial ? chunk. isInitial() : chunk.initial`

a-x-:
Thank you,

As I understand isInitial might be a replacement for the `chunk.entrypoints`.

But I'm not sure to make this new logic default for all users in the next minor version.

mastilver:
I doubt it can be considered as a breaking change,
to me it's definitely a fix, so it should land on a patch (BTW, some tests would be great! ;) )

lib/plugin.js
```diff
@@ -42,24 +42,26 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
-    _.merge(manifest, compilation.chunks.reduce(function(memo, chunk) {
-      var chunkName = chunk.name ? chunk.name.replace(this.opts.stripSrc, '') : null;
-
-      // Map original chunk name to output files.
-      // For nameless chunks, just map the files directly.
-      return chunk.files.reduce(function(memo, file) {
-        // Don't add hot updates to manifest
-        if (file.indexOf('hot-update') >= 0) {
+    _.merge(manifest, compilation.chunks
+      .filter(chunk => chunk => chunk.isInitial ? chunk.isInitial() : chunk.initial)
```
mastilver: extra `chunk => `

a-x-:
done

Ianfeather:
@mastilver Hi! First of all thanks for your work on this project, we use it in the asset-pipeline at BuzzFeed. 

Unfortunately this change broke one of our applications this morning as it was relying on the dynamic chunks being included in the asset-manifest. We use these manifest entries to preload bundles ahead of time with `link[rel=preload]`. We can probably find some alternative but it would be great if you would consider having this as an option. I think this is a very legitimate use case.

a-x-:
Hey, we wanted to make it optional initially 😂

mastilver:
@Ianfeather Damn, I was really hoping for my first release to go smoothly... :/

I will revert that change for now, we can think for another solution later

a-x-:
@Ianfeather, you should use `chunk-manifest-webpack-plugin`

my position, this change should be default but major
@mastilver, what about you?

mastilver:
@Ianfeather `webpack-manifest-plugin@1.0.2` released :)

mastilver:
@a-x- I thing the right path to take is: by default include every single possible assets into the manifest and be able to have some filtering, I will put my thought down on an issue over the week-end, I will be happy to get your thoughts on that ;)

a-x-:
ok, I can restore initial flagged functionality

mastilver:
@a-x- I want to work on something more modular, it will solve your issue and few others

In the meantime, I would love to get some help on #51 as I won't be merging anything else until that's done. that way I'm sure we don't introduce new breaking changes

Ianfeather:
@mastilver Thanks for your quick turnaround on this! It's much appreciated.

@a-x- thanks for the link to that project. That will also help us remove our webpack-runtime script too

lib/plugin.js
```diff
@@ -42,24 +42,26 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
-      var chunkName = chunk.name ? chunk.name.replace(this.opts.stripSrc, '') : null;
-
-      // Map original chunk name to output files.
-      // For nameless chunks, just map the files directly.
-      return chunk.files.reduce(function(memo, file) {
-        // Don't add hot updates to manifest
-        if (file.indexOf('hot-update') >= 0) {
+    _.merge(manifest, compilation.chunks
+      .filter(chunk => chunk.isInitial ? chunk.isInitial() : chunk.initial)
+      .reduce(function(memo, chunk) {
```
zuzusik: It looks like if we use `(memo, chunk) => {...}` we get the right context and we can skip `.bind(this)`

lib/plugin.js
```diff
@@ -42,24 +42,26 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
+        var chunkName = chunk.name ? chunk.name.replace(this.opts.stripSrc, '') : null;
+
+        // Map original chunk name to output files.
+        // For nameless chunks, just map the files directly.
+        return chunk.files.reduce(function(memo, file) {
```
zuzusik: same as https://github.com/danethurber/webpack-manifest-plugin/pull/48/files#r126661193, but will work only in combination with the previous (e.g. if both functions are arrow)

lib/plugin.js
mastilver: @zuzusik You are right, but it can't happens before  #53 
I did the mistake of merging this PR without caring for older node versions
I added tests to prevent this kind of mistake in the future

zuzusik: cool, got it

Issue #33, opened 2017-01-18
an option to create the manifest json to specified directory


tomwang1013:
Now the manifest.json is created under `output.path` which is set in webpack.config.js. But i want it to be in other directory.

btamas:
It's working for me if you define filname as a path of output.
```
...
new ManifestPlugin({
    fileName : path.join(__dirname, '..', 'rev-manifest.json'),
    basePath : '/js/',
  })
...
```

tomwang1013:
@btamas this didn't work for me. Say `output.path` is `public/assets`, the absolute path of my application is `/home/wangxiantong/learning/mysite`, the result is:

![default](https://cloud.githubusercontent.com/assets/1826003/22912810/50142e4a-f2a2-11e6-8162-bfd4614be9ba.png)

which is terribly incorrect.

dcurletti:
I second this problem.  I too am having issues with wanting to write to a specific directory.

arthens:
Using a relative path as a file name worked for me

```
new ManifestPlugin({
     fileName : '../../manifest.json',
}),
```

but it feels very hacky and I agree that it should be its own configuration option.

zuzusik:
btw, this got broken somewhere after `1.0.0`
with previous version we had `fileName: './rev-manifest-webpack.json',` and the file was placed to the folder in which webpack config is placed where with latest version it gets placed in the `dist` folder - where we do the webpack output

this is not good - I don't want my dist folder to have this file - it is not needed in dist

mastilver:
I agree with you, I think `manifest.json` path should be relative to `cwd` or `context`

That causing the issue, is this: https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L97-L104

I think it should be dropped, and the plugin should write to disk by himself, no rely on webpack to do it

I also believe that it will fix: #21 

Thoughts?

mastilver:
@btamas 's solution should work for now

for `v2`, this plugin will make sure that `fileName` is an absolute path

zuzusik:
> That causing the issue, is this: https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L97-L104
> I think it should be dropped, and the plugin should write to disk by himself, no rely on webpack to do it

@mastilver if we just drop these lines you reference we bring back other bugs, look at the description of the PR where those lines were added: https://github.com/danethurber/webpack-manifest-plugin/pull/6

[...]

Issue #49, opened 2017-07-03
Looking for maintainers


mastilver:
@danethurber just added me as a collaborators to this project

I think it will be better if I'm not the only one making the decisions on this project! :)

So if you are ready to spend some time maintaining this project, ping me here

ping @jordansexton @gmac @andreicek @tikotzky @martynovs @nickescallon @modosc  @iliakan @a-x- @davidmpaz @Lobstrosity @weaverryan @craigmichaelmartin @wtgtybhertgeghgtwtg @orbiteleven

wtgtybhertgeghgtwtg:
I have a few ideas that could help, sure, but how did my name come up?  I don't think I've ever touched this project.
Did you confuse my name for somebody else's?  Happens all the time.

craigmichaelmartin:
https://github.com/danethurber/webpack-manifest-plugin/pull/32

mastilver:
@wtgtybhertgeghgtwtg I pinged everybody who contributed / opened a PR

zuzusik:
I can definitely spend some time reviewing PRs, but as I am not an expert in how `WebPack` works under the hood this would be mostly general logic checking/analyzing + code style reviews

a-x-:
+1, Me too
Let's try

mastilver:
I'm quite familiar with `webpack`, not that much with the history of `webpack-manifest-plugin`
So I guess with the three of us we have a good mixed team ;)

I will ask Dane to add you as Collaborator, the only rule is don't push to `master` changes impacting the plugin's logic

mastilver:
I'm closing this now, but feel free to reach out here if you are willing to help ;)

gmac:
I don't mind helping out

mastilver:
@gmac definitely, I think you can help on 2 and 3 of #51

mastilver:
@gmac @a-x- @zuzusik I created a slack channel here: https://webpack.slack.com/messages/C6CCZ0DSS

Anyone else welcome of course ;)

a-x-:
Can you create an invite to invntrm@gmail.com?

mastilver:
I asked @d3viant0ne to add all of you and @danethurber on slack

mastilver:
Alright, so I'm created a gitter instead: https://gitter.im/webpack-manifest-plugin#

PR #56, opened 2017-07-09
fix: make sure manifest writes don't overlap

### LINKED ISSUE #21: Mangled output
In multi compiler mode with a shared cache, I am getting the following mangled output.

```
{
  "common_en-US.js": "common_en-US-cbe78523083d0159240f.js",
  "common_en-US.js.map": "common_en-US-cbe78523083d0159240f.js.map"
}  "common_fr-CA.js": "common_fr-CA-2cf11bd51dc92d8b0ee2.js",
  "common_fr-CA.js.map": "common_fr-CA-2cf11bd51dc92d8b0ee2.js.map"
}
```

This is inconsistent, sometimes the file is correct, sometimes it is as above

mastilver:
closes #21

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -13,7 +13,8 @@ function ManifestPlugin(opts) {
     stripSrc: null,
     transformExtensions: /^(gz|map)$/i,
     writeToFileEmit: false,
-    cache: null
+    cache: null, // TODO: Remove `cache` in next major release in favour of `seed`.
+    seed: null
   }, opts || {});
 }
 
@@ -29,7 +30,7 @@ ManifestPlugin.prototype.getFileType = function(str) {
 
 ManifestPlugin.prototype.apply = function(compiler) {
   var outputName = this.opts.fileName;
-  var cache = this.opts.cache || {};
+  var seed = this.opts.seed || this.opts.cache || {};
   var moduleAssets = {};
 
   compiler.plugin("compilation", function (compilation) {
@@ -92,10 +93,10 @@ ManifestPlugin.prototype.apply = function(compiler) {
     }
 
     Object.keys(manifest).sort().forEach(function(key) {
-      cache[key] = manifest[key];
+      seed[key] = manifest[key];
     });
 
-    var json = JSON.stringify(cache, null, 2);
+    var json = JSON.stringify(seed, null, 2);
 
     var outputFolder = compilation.options.output.path;
     var outputFile = path.join(outputFolder, this.opts.fileName);
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -99,7 +99,7 @@ ManifestPlugin.prototype.apply = function(compiler) {
     var json = JSON.stringify(seed, null, 2);
 
     var outputFolder = compilation.options.output.path;
-    var outputFile = path.join(outputFolder, this.opts.fileName);
+    var outputFile = path.resolve(compilation.options.output.path, this.opts.fileName);
 
     if (this.opts.writeToFileEmit) {
       fse.outputFileSync(outputFile, json);
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -1,6 +1,9 @@
 var path = require('path');
 var fse = require('fs-extra');
 var _ = require('lodash');
+var mutexify = require('mutexify');
+
+var lock = mutexify();
 
 function ManifestPlugin(opts) {
   this.opts = _.assign({
@@ -94,23 +97,23 @@ ManifestPlugin.prototype.apply = function(compiler) {
 
     var json = JSON.stringify(cache, null, 2);
 
-    compilation.assets[outputName] = {
-      source: function() {
-        return json;
-      },
-      size: function() {
-        return json.length;
-      }
-    };
+    var outputFolder = compilation.options.output.path;
+    var outputFile = path.join(outputFolder, this.opts.fileName);
 
     if (this.opts.writeToFileEmit) {
-      var outputFolder = compilation.options.output.path;
-      var outputFile = path.join(outputFolder, this.opts.fileName);
-
       fse.outputFileSync(outputFile, json);
     }
 
-    compileCallback();
+    compiler.outputFileSystem.mkdirp(path.dirname(outputFile), function(err) {
+      if (err) return compileCallback(err);
+
+      lock(function(release) {
+        compiler.outputFileSystem.writeFile(outputFile, json, function(err) {
+          release();
+          compileCallback(err);
+        });
+      });
+    });
   }.bind(this));
 };
 
--- a/spec/plugin.integration.spec.js
+++ b/spec/plugin.integration.spec.js
@@ -1,4 +1,4 @@
-var fs = require('fs');
+var fse = require('fs-extra');
 var path = require('path');
 
 var _ = require('lodash');
@@ -21,7 +21,10 @@ function webpackCompile(config, compilerOps, cb) {
 
   _.assign(compiler, compilerOps);
 
-  compiler.run(function(err, stats){
+  compiler.watch({
+    aggregateTimeout: 300,
+    poll: true
+  }, function(err, stats){
     expect(err).toBeFalsy();
     expect(stats.hasErrors()).toBe(false);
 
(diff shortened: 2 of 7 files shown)


zuzusik:
The main question here - does it really fix #21? I don't see an appropriate test case added here. If it is hard to reproduce maybe we should ask reporters of #21 to test this fix with their local setup?

If it doesn't fix #21 then I don't see why it should be merged. As a step for #33 as supposed here
 https://github.com/danethurber/webpack-manifest-plugin/issues/33#issuecomment-314063783 ?
Don't think we need this step - we better fix #33 with one monolithic PR

lib/plugin.js
```diff
@@ -94,20 +94,16 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
-
-    if (this.opts.writeToFileEmit) {
-      var outputFolder = compilation.options.output.path;
-      var outputFile = path.join(outputFolder, this.opts.fileName);
+    var outputFolder = compilation.options.output.path;
+    var outputFile = path.join(outputFolder, this.opts.fileName);
 
+    const shouldUseNodeFs = compiler.outputFileSystem.constructor.name !== 'MemoryFileSystem' ||
+                                      this.opts.writeToFileEmit;
+    if (shouldUseNodeFs) {
```
zuzusik: As far as I see, the logic used here breaks what was achieved with https://github.com/danethurber/webpack-manifest-plugin/pull/28, as it assumes the file should be output to both memory and file system - now when `writeToFileEmit` is set ti true, it will be output only to file system, and not to memory.

mastilver: 👍  will fix that

mastilver:
@zuzusik Thank you for reviewing this PR :)

Did you see #49, I think you could be a good help for me

I think I know what's causing the issue, I will try to write a test that reproduce it

lib/plugin.js
```diff
@@ -94,20 +94,16 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
-    };
-
-    if (this.opts.writeToFileEmit) {
-      var outputFolder = compilation.options.output.path;
-      var outputFile = path.join(outputFolder, this.opts.fileName);
+    var outputFolder = compilation.options.output.path;
+    var outputFile = path.join(outputFolder, this.opts.fileName);
 
+    const shouldUseNodeFs = compiler.outputFileSystem.constructor.name !== 'MemoryFileSystem' ||
+                                      this.opts.writeToFileEmit;
```
zuzusik: this indentation looks broken, I think we should stick to 2 spaces everywhere, also I think it is little bit more readable if we put `||` on the new line:
```js
const shouldUseNodeFs = compiler.outputFileSystem.constructor.name !== 'MemoryFileSystem'
  || this.opts.writeToFileEmit;
```

spec/plugin.integration.spec.js
```diff
@@ -0,0 +1,53 @@
+var fs = require('fs');
+var path = require('path');
+
+var _ = require('lodash');
+var webpack = require('webpack');
+
+var ManifestPlugin = require('../index.js');
+
+function webpackConfig (webpackOpts, opts) {
```
zuzusik: let's be consistent with style here - we should stick to `function<name>(...) {}` style as it is used all over the project

I mean - there shouldn't be a space before `(`

Also it should be fixed in other places in this PR

lib/plugin.js
```diff
@@ -94,20 +94,16 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
-      }
-    };
-
-    if (this.opts.writeToFileEmit) {
-      var outputFolder = compilation.options.output.path;
-      var outputFile = path.join(outputFolder, this.opts.fileName);
+    var outputFolder = compilation.options.output.path;
+    var outputFile = path.join(outputFolder, this.opts.fileName);
 
+    const shouldUseNodeFs = compiler.outputFileSystem.constructor.name !== 'MemoryFileSystem' ||
```
zuzusik: I know that `const` is much more appropriate here, but this file uses `var` all over the place (and even 2 previous lines use `var`)

So I guess we should stick to `var` and if we want to go with `const` we should create separate PR in which we do refactor all `var` to `const`

spec/plugin.spec.js
```diff
@@ -62,7 +62,7 @@ function webpackCompile(webpackOpts, opts, cb) {
   var fs = compiler.outputFileSystem = new MemoryFileSystem();
 
   compiler.run(function(err, stats){
-    var manifestFile = JSON.parse( fs.readFileSync(manifestPath).toString() );
+    var manifestFile = JSON.parse(fs.readFileSync(manifestPath).toString());
```
zuzusik: I'd remove this changes as they are completely unrelated to this PR

zuzusik:
Also added few style related comments here

lib/plugin.js
zuzusik: Also, as far as I see we do rely on internal implementation of [`memory-fs`](https://github.com/webpack/memory-fs) here. Not sure if it is a good idea - the class name can be changed at any time without implicating the API, so it wouldn't be a reason for new major version release.

zuzusik: And another question here - what if someone provides custom memory file system which is implemented in the other way?

lib/plugin.js
```diff
@@ -94,20 +94,16 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
+                                      this.opts.writeToFileEmit;
+    if (shouldUseNodeFs) {
       fse.outputFileSync(outputFile, json);
+    } else {
+      compiler.outputFileSystem.mkdirpSync(path.dirname(outputFile));
```
zuzusik: I wonder why `mkdirpSync` and not `mkdirp`?
Webpack docs only mention `mkdirp` [here](https://webpack.js.org/api/node/#custom-file-systems) so it might blow up for some custom FS that provides only `mkdirp` but doesn't provide `mkdirpSync`

zuzusik:
One serious concern here: https://github.com/danethurber/webpack-manifest-plugin/pull/56/files#r126405642

lib/plugin.js
mastilver: I agree it's ugly but that the only way I could find to check if it's really `MemoryFileSystem`, if you have a better solution feel free to share it ;)

I'm note the only one doing it: https://github.com/gajus/write-file-webpack-plugin/blob/51167a8cf33ee0885805abd1fb85c8306f0eda27/src/index.js#L19

lib/plugin.js
mastilver: Should be `mkdirp.sync`

lib/plugin.js
zuzusik: What do you think about this proposal:

```js
var outputFolder = compilation.options.output.path;
var outputFile = path.join(outputFolder, this.opts.fileName);

compiler.outputFileSystem.mkdirp(path.dirname(outputFile));
compiler.outputFileSystem.writeFileSync(outputFile, json);  

if ( this.opts.writeToFileEmit) {
  fse.outputFileSync(outputFile, json);
}
```

mastilver: That would only work when using `memory-fs`, see: https://github.com/webpack/webpack/blob/master/lib/node/NodeOutputFileSystem.js#L17
It doesn't have `writeFileSync`

mastilver: this should work thought: 

```
compiler.outputFileSystem.mkdirp(path.dirname(outputFile));
if (isMemoryFs) {
  compiler.outputFileSystem.writeFileSync(outputFile, json);  
}

if (!isMemoryFs || this.opts.writeToFileEmit) {
  fse.outputFileSync(outputFile, json);
}
```

zuzusik: Your code looks good, should work as expected 👍 I only assume it should use `mkdirp.sync`

What I want to achieve here is that we can skip `compiler.outputFileSystem.constructor.name !== 'MemoryFileSystem'` check

Overall impression - it is weird that different FSs used in WebPack have different sets of methods.

mastilver: > I only assume it should use mkdirp.sync

Damn, I updated it :)

mastilver:
@zuzusik PR updated, do you mind having another look

mastilver:
@zuzusik I'm not relying anymore or `memory-fs`logic

Let me know what you think!

Issue #55, opened 2017-07-08
Add reduce, initValue, map, filter options


mastilver:
- [x] filter (#66)
- [x] map
- [x] reduce
- [x] seed (#57)

-----

fix: #48 
```js
filter: ({chunk, file}) => {
  return chunk.isInitial();
}
```

fix: #25 

```js
initValue: {

} // will be used as the initValue for reduce, default to `{}`
```

fix #35 

```js
map: ({chunk, file}) => {
  return {
    file,
    hash: chunk.source.integrity
  };
}
```

```js
reduce: (manifest, {chunk, file}) => {
  return manifest;
}
```

-----
`map` and `reduce` can't be used at the same time

nathanbirrell:
Hey @mastilver I've added the initValue on my fork. Created PR https://github.com/danethurber/webpack-manifest-plugin/pull/57

mastilver:
@a-x- @zuzusik @gmac Once unit tests are close to 💯  I would like to start working on that
After re-thinking it I'm not sure if `reduce` is necessary; `map` and `filter` will probably be enough 

Thoughts?

weaverryan:
Just dropping a note that this would also solve #37, where we want to be able to tweak (prefix) both the keys and values of the manifest. This would be a much more flexible way of handling it than the one-off option I proposed there.

a-x-:
`initValue` name is too abstract as I see.

PR #57, opened 2017-07-11
Deprecate cache parameter in favour of seed

### LINKED ISSUE #25: Define other manifest fields
Is there some way to predefine other manifest fields, e.g. name, short_name, icons, start_url, background_color, display etc. ?

nathanbirrell:
Functionality for `cache` stays the same, the more generic name is better suited for when your app needs to set [custom keys](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json).

requested in #55

-----
closes #25


lib/plugin.js
```diff
@@ -88,6 +90,9 @@ ManifestPlugin.prototype.apply = function(compiler) {
       }.bind(this), {});
     }
 
+    // Add custom values last, so they don't get the basePath prefixed to them
+    _.merge(manifest, initValue);
```
mastilver: use `manifest = _.assign({}, initValue, manifest)`

spec/plugin.spec.js
```diff
@@ -214,6 +214,58 @@ describe('ManifestPlugin', function() {
       });
     });
```
mastilver: Can you add a test to prove that `initValue` is being overwritten

mastilver:
@a-x- @zuzusik @gmac Does `initValue` make sense to you (I had no idea how to name it so I just looked at the doc for `Array.reduce` :D )

nathanbirrell:
Hey @gmac, I agree.

I've updated the PR, unit tests and readme to reflect the move to `seed` (mentioned [here](https://github.com/danethurber/webpack-manifest-plugin/issues/25#issuecomment-314487592)).

Cheers

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webpack-manifest-plugin",
-  "version": "1.1.2",
+  "version": "1.1.4",
```
mastilver: Please leave the version as is :)

mastilver:
@gmac are you ok with those changes? 
Or shall we output a deprecation warning using `util.deprecate`?

package.json
nathanbirrell: Sure, done

spec/plugin.spec.js
```diff
@@ -214,7 +214,85 @@ describe('ManifestPlugin', function() {
...
+        expect(manifest).toEqual({
+          'one.js': 'one.js',
+          'test1': 'test2'
+        });
+
+        done();
+      });
+    });
+
+    it('seed attributes do not get prefixed with basePath', function(done) {
```
zuzusik: I think we should go here with `does something` style - `does not prefix seed attributes with basePath`

spec/plugin.spec.js
```diff
@@ -214,7 +214,85 @@ describe('ManifestPlugin', function() {
...
+          'test1': 'test2'
+        });
+
+        done();
+      });
+    });
+
     it('combines manifests of multiple compilations', function(done) {
+      var seed = {};
+      webpackCompile([{
```
zuzusik: 1. do we need this `var` at all?
2. shouldn't we place a new line after `var`?

spec/plugin.spec.js
nathanbirrell: Nice one, done.

README.md
```diff
@@ -52,4 +55,4 @@ new ManifestPlugin({
 * `publicPath`: A path prefix used only on output files, similar to Webpack's  [output.publicPath](https://github.com/webpack/docs/wiki/configuration#outputpublicpath). Ignored if `basePath` was also provided.
 * `stripSrc`: removes unwanted strings from source filenames (string or regexp)
 * `writeToFileEmit`: If set to `true` will emit to build folder and memory in combination with `webpack-dev-server`
-* `cache`: In [multi-compiler mode](https://github.com/webpack/webpack/tree/master/examples/multi-compiler) webpack will overwrite the manifest on each compilation. Passing a shared `{}` as the `cache` option into each compilation's ManifestPlugin will combine the manifest between compilations.
+* `seed`: A cache of key/value pairs to used to seed the manifest. This may include a set of [custom key/value](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json) pairs to include in your manifest, or may be used to combine manifests across compilations in [multi-compiler mode](https://github.com/webpack/webpack/tree/master/examples/multi-compiler).
```
zuzusik: Should we add here the information how exactly it should be used in multi-compiler-mode?

Maybe I am missing something, but for me it is not really clear that I need to specify it as an empty object to do that - I cleared this out from tests, but who is going to read tests when he needs info?)

gmac: How about just adding: `To combine manifests, pass a shared seed object to each compiler's plugin instance`.

zuzusik: > `To combine manifests, pass a shared seed object to each compiler's plugin instance.`

this can be somehow confusing for those not knowing how it all works under the hood - the first question will be  `what is compiler plugin instance?`

I assume adding `: just pass it as empty object: 'seed: {}'` before final `.`

zuzusik:
Thinking about further improvements here: is there a way to detect that we are in multi-compiler mode inside of a plugin and then switch to seeding mode?

This would be really great and convenient - no need to specify seed as `{}` - it will just work out of the box.

README.md
gmac: I find `just pass it as empty object: 'seed: {}'` more confusing than helpful. It doesn't make that point that a shared singleton object must be passed to multiple plugin instances to combine their manifests. I amend my earlier text for clarity:

```To combine manifests, pass a shared seed object to each compiler's ManifestPlugin instance.```

If you're doing multi-compiler architecture, that should make sense. If you're not, you don't need the information anyway.

gmac:
Sure, it's possible to detect multi-compiler mode... you just look at the number of compilers configured. Off the top of my head, I think that's managed by an array called `compilers` on the root Webpack compiler object.

This sounds like a potentially useful feature, although beyond the scope of this PR. There's definitely some nuances there that would need to be solved for. The way I'd see it working would be to pass an option like `universalManifest: true`. When enabled, a plugin instance would configure around an internal singleton and mixin any provided seeds onto that. Otherwise, it would configure around the provided seed object itself. The nuance would be in tracking the right object identity.

nathanbirrell:
Thanks all. Updated the readme with @gmac's suggestion 👍🏼

mastilver:
👍  I will release that tonight

PR #58, opened 2017-07-11
Removing lodash dependency

### LINKED ISSUE #53: Deprecate node 0.10 and 0.12
- [x] `travis.yml`: https://github.com/danethurber/webpack-manifest-plugin/blob/master/.travis.yml#L6-L7 https://github.com/danethurber/webpack-manifest-plugin/blob/master/.travis.yml#L19
- [x]  test: https://github.com/danethurber/webpack-manifest-plugin/blob/master/spec/plugin.integration.spec.js#L97-L98

gmac:
With this plugin gaining traction as a significant webpack/webpacker resource, seems like time to do some housekeeping that trims it down. This removes Lodash.

-----
closes #53


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webpack-manifest-plugin",
-  "version": "1.1.2",
```
mastilver: don't update version

mastilver:
I think this should be done as part of #53 

Let's leave it open for now, we will merge it when we are ready to release `v2`! :)

package.json
```diff
@@ -9,7 +9,7 @@
   "author": "Dane Thurber <dane.thurber@gmail.com>",
   "license": "MIT",
   "engines": {
-    "node": ">=0.10"
+    "node": ">=4.3.0"
```
mastilver: `>=4`

You might also want to remove old node version from `.travis.yml`

gmac: Ah, thank you. I missed the Travis file.

mastilver:
@gmac I already deprecated node@0.10 and node@0.12, but I didn't do any code change
So your PR is still valid and can be merged after 2.x as it won't be a breaking change (or before if you have time to work on it ;)

andriijas:
@mastilver merge? :)

Issue #51, opened 2017-07-07
Improve tests coverage


mastilver:
Those are not covered by tests:
- [x] https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L34-L37 (#59)
- [x] https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L52-L54
- [x] https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L58 (#59)
- [x] https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L68-L70 (#59)
- [x] https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L108-L113 (#56)

a-x-:
I'm starting look at #L34-L37

mastilver:
I'm working on L52-L54
WIP: https://github.com/danethurber/webpack-manifest-plugin/tree/test-update-hot

PR #59, opened 2017-07-11
add test: outputs a manifest of no-js file

a-x-:
https://github.com/danethurber/webpack-manifest-plugin/issues/51
https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L34-L37


package.json
```diff
@@ -18,6 +18,7 @@
     "codecov": "^2.2.0",
     "css-loader": "^0.9.1",
     "extract-text-webpack-plugin": "^0.5.0",
+    "file-loader": "^0.10.0",
```
mastilver: use 0.9.0 for older node compatibility

mastilver:
:+1: Awesome work! you ticked 3 out of 5 boxes :)

package.json
a-x-: done
thank you!

PR #61, opened 2017-07-20
Allow publicPath and basePath together

gmac:
Another adjustment for the v.2 overhaul: this cleans up the roles and responsibilities of `publicPath` and `basePath`. Two specific changes:

1. The public and base path options are no longer mutually exclusive; they may be used together. This alleviates each setting's individual limitations.
2. Base path gets applies _after_ public path so that the output files may be relativized, and then the whole build. This feels a lot more predictable than the two settings toggling each other.

This technically represents a breaking change, although there is low likelihood of consumer issues (given that there is literally _no_ reason to include `publicPath` in the presence of `basePath` at the moment). If consumers do currently define both options and their build breaks in v2, the fix is simple: just delete `publicPath`.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -75,19 +75,22 @@ ManifestPlugin.prototype.apply = function(compiler) {
       return memo;
     }, {}));
 
-    // Append optional basepath onto all references.
-    // This allows output path to be reflected in the manifest.
+    // Prepend an optional public path onto all manifest values,
+    // similar to Webpack's `output.publicPath` option. Example:
+    // {"foo/bar.js":"foo/bar.123.js"} becomes {"foo/bar.js":"public/foo/bar.123.js"}
+    if (this.opts.publicPath) {
+      manifest = _.reduce(manifest, function(memo, value, key) {
+        memo[key] = resolvePath(this.opts.publicPath, value);
+        return memo;
+      }.bind(this), {});
+    }
+
+    // Prepend an optional base path onto all manifest keys AND values.
+    // This allows Webpack's `output.path` to be reflected in the manifest. Example:
+    // {"foo/bar.js":"foo/bar.123.js"} becomes {"path/to/foo/bar.js":"path/to/foo/bar.123.js"}
     if (this.opts.basePath) {
       manifest = _.reduce(manifest, function(memo, value, key) {
         memo[resolvePath(this.opts.basePath, key)] = resolvePath(this.opts.basePath, value);
-        return memo;
-      }.bind(this), {});
-    } else if (this.opts.publicPath) {
-      // Similar to basePath but only affects the value (similar to how
-      // output.publicPath turns require('foo/bar') into '/public/foo/bar', see
-      // https://github.com/webpack/docs/wiki/configuration#outputpublicpath
-      manifest = _.reduce(manifest, function(memo, value, key) {
-        memo[key] = resolvePath(this.opts.publicPath, value);
         return memo;
       }.bind(this), {});
     }
--- a/README.md
+++ b/README.md
@@ -48,8 +48,8 @@ new ManifestPlugin({
 **Options:**
 
 * `fileName`: The manifest filename in your output directory (`manifest.json` by default).
-* `basePath`: A path prefix for all file references. Useful for including your output path in the manifest.
-* `publicPath`: A path prefix used only on output files, similar to Webpack's  [output.publicPath](https://github.com/webpack/docs/wiki/configuration#outputpublicpath). Ignored if `basePath` was also provided.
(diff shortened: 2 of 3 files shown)


mastilver:
👍  I think it make sense

weaverryan:
Thanks for this @gmac!

First... I apologize for being difficult! I'm realizing that we must be doing something in the [Webpack Encore](https://github.com/symfony/webpack-encore) library that is a bit non-traditional with our manifest :/. This PR won't yet work for us. 

To use your example, suppose the manifest starts like this:

```json
{"foo/bar.js":"foo/bar.123.js"}
```

If the output path is some `build/` dir, then we could use `basePath` to transform to this:

```json
{"build/foo/bar.js":"build/foo/bar.123.js"}
```

So far, we're good! But, we also allow people to set their public path to a CDN (e.g. `https://coolcdn.com/build`). Technically, that CDN doesn't need to be reflected in the manifest values... but it's *really* convenient for our users to be able to transform the manifest to this:

```json
{"build/foo/bar.js":"https://coolcdn.com/build/foo/bar.123.js"}
```

And this is the part that's not possible yet. With this PR, the we would set the `publicPath` to the CDN and the `basePath` to `build/`, but would end up (I think) with a manifest like this:

```json
{"build/foo/bar.js":"build/https://coolcdn.com/build/foo/bar.123.js"}
```

Does that make sense? I think we're coming from a community with a slightly different paradigm, so I think I'm having trouble explaining myself :). We read the `manifest.json` file in server-side code to convert from `build/foo/bar.json` to the real value that should be used in the `<script>` tag. Having the CDN there means the user can change their `publicPath` to a CDN in webpack... and their server-side code automatically starts using it without any changes (server-side templates always reference `build/foo/bar.json`).

Thanks!

mastilver:
@gmac I believe @weaverryan is right

`basePath` should be for the key 
`publicPath` for the value

his issue was because the code is currently like that

```js
if (publicPath) {

} else if (basePath) {

}
``` 

It would only pick up `publicPath` or `basePath`, not both

lib/plugin.js
```diff
@@ -75,19 +75,22 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
-      // https://github.com/webpack/docs/wiki/configuration#outputpublicpath
+    }
+
+    // Prepend an optional base path onto all manifest keys AND values.
+    // This allows Webpack's `output.path` to be reflected in the manifest. Example:
+    // {"foo/bar.js":"foo/bar.123.js"} becomes {"path/to/foo/bar.js":"path/to/foo/bar.123.js"}
+    if (this.opts.basePath) {
       manifest = _.reduce(manifest, function(memo, value, key) {
-        memo[key] = resolvePath(this.opts.publicPath, value);
+        memo[resolvePath(this.opts.basePath, key)] = resolvePath(this.opts.basePath, value);
```
mastilver: `basePath` is only for the key

lib/plugin.js
```diff
@@ -75,19 +75,22 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
-    // Append optional basepath onto all references.
-    // This allows output path to be reflected in the manifest.
-    if (this.opts.basePath) {
+    // Prepend an optional public path onto all manifest values,
+    // similar to Webpack's `output.publicPath` option. Example:
+    // {"foo/bar.js":"foo/bar.123.js"} becomes {"foo/bar.js":"public/foo/bar.123.js"}
+    if (this.opts.publicPath) {
       manifest = _.reduce(manifest, function(memo, value, key) {
-        memo[resolvePath(this.opts.basePath, key)] = resolvePath(this.opts.basePath, value);
+        memo[key] = resolvePath(this.opts.publicPath, value);
```
mastilver: I reverted the commit that introduced `resolvePath()`

mastilver:
I found this in the readme:

> Ignored if basePath was also provided.

Which I think is wrong, no?

weaverryan:
@mastilver are you talking about the readme on master on this PR? I wasn't sure what you meant :).

 That note is accurate for the behavior on master, but was removed here. That's actually the behavior on master I'm hoping we can change :).

mastilver:
@weaverryan I must be a bit tired :D 

You are right, this PR, is fixing the fact that `publicPath` is being ignored ;)

lib/plugin.js
mastilver: I just realized, I was wrong about that
lgtm then! :)

mastilver: I understand that the current behavior of `basePath` is acting on both key and value.
I believe it should only modify the key. Thoughts ?

mastilver:
@gmac I think it should follow #69 instead

mastilver:
closing is flavour of #74

PR #66, opened 2017-07-28
feat: add filter, map, reduce options

### LINKED ISSUE #35: output file hash to manifest
What do you guys think about adding hash information to manifest file so we can make use of SRI?
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity?
> Subresource Integrity (SRI) is a security feature that enables browsers to verify that files they fetch (for example, from a CDN) are delivered without unexpected manipulation. It works by allowing you to provide a cryptographic hash that a fetched file must match.

### LINKED ISSUE #46: Do not add code splitting bundles into manifest

### LINKED ISSUE #55: Add reduce, initValue, map, filter options

mastilver:
closes #46 
closes #55 
closes #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,8 +114,6 @@ Type: `function`
 
 Create the manifest. It must return an `Object`.  The default one will take the name of the chunk as a key and the path of the file as the value.
 
-> NOTE: Using `reduce` will ignore `basePath` and `publicPath` options
-
 
 ## Hooks Options
 
--- a/README.md
+++ b/README.md
@@ -99,20 +99,22 @@ A cache of key/value pairs to used to seed the manifest. This may include a set 
 
 Type: `function`
 
-Filter out files.
+Filter out files. [Check parameters in details](#hooks-options)
 
 
 ### `map`
 
 Type: `function`
 
-Modify files details before the manifest is created.
+Modify files details before the manifest is created. [Check parameters in details](#hooks-options)
+
 
 ### `reduce`
 
-Type: `function`
+Type: `function`<br>
+Default: `(manifest, {name, path}) => ({...manifest, [name]: path})`
 
-Create the manifest. It must return an `Object`.  The default one will take the name of the chunk as a key and the path of the file as the value.
+Create the manifest. It must return an `Object`. Use the `seed` options to populate `manifest`. [Check parameters in details](#hooks-options)
 
 
 ## Hooks Options
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ new ManifestPlugin({
 })
 ```
 
+
 ## Options:
 
 ### `fileName`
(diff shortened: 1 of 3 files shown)


spec/plugin.spec.js
```diff
@@ -393,4 +393,29 @@ describe('ManifestPlugin', function() {
...
+        entry: {
+          nameless: './fixtures/nameless.js'
+        },
+        output: {
+          filename: '[name].[hash].js'
+        }
+      }, {
+        manifestOptions: {
+          filter: function(entry) {
+            return entry.isInitial;
```
mastilver: @a-x- Does that looks good to you?

spec/plugin.spec.js
```diff
@@ -393,4 +393,85 @@ describe('ManifestPlugin', function() {
...
+        context: __dirname,
+        entry: './fixtures/file.js',
+        output: {
+          filename: '[name].js'
+        }
+      }, {
+        manifestOptions: {
+          reduce: function (manifest, file) {
+            manifest[file.name] = {
+              file:file.path,
```
a-x-: `:<space>`

README.md
```diff
@@ -94,6 +95,68 @@ Default: `{}`
...
+
+Type: `function`
+
+Modify files details before the manifest is created.
+
+### `reduce`
+
+Type: `function`
+
+Create the manifest. It must return an `Object`.  The default one will take the name of the chunk as a key and the path of the file as the value.
```
a-x-: As I understand, we should note here: «custom reduce opt function replaces that default `reduce` logic:
```js
function (manifest, file) {
        manifest[file.name] = file.path;
        return manifest;
}
```

spec/plugin.spec.js
a-x-: are we need eslint?

mastilver: We definitely need to setup eslint as some point

I personally have a preference for https://github.com/sindresorhus/xo , but really I don't mind ;)

README.md
```diff
@@ -94,6 +95,68 @@ Default: `{}`
 
 A cache of key/value pairs to used to seed the manifest. This may include a set of [custom key/value](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json) pairs to include in your manifest, or may be used to combine manifests across compilations in [multi-compiler mode](https://github.com/webpack/webpack/tree/master/examples/multi-compiler). To combine manifests, pass a shared seed object to each compiler's ManifestPlugin instance.
 
+### `filter`
+
+Type: `function`
+
+Filter out files.
```
mastilver: I don't know, I don't really want to pollute the readme too much...

lib/plugin.js
```diff
@@ -41,36 +44,73 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
+
+
+    var manifest;
+    if (this.opts.reduce) {
+      manifest = files.reduce(this.opts.reduce, {});
+    } else {
+      manifest = files.reduce(function (manifest, file) {
+        manifest[file.name] = file.path;
+        return manifest;
+      }, {});
```
mastilver: use `seed` instead of `{}`

spec/plugin.spec.js
a-x-: is xo like [prettier](https://github.com/prettier/prettier)?

PR #67, opened 2017-07-31
chore(xo): setup linter

mastilver:
Once v1.3 is release, and #58 is merged; this should be merged with linting fixes

Any additional rules, you want in / disable?

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "webpack plugin for generating asset manifests",
   "main": "index.js",
   "scripts": {
-    "test": "nyc jasmine"
+    "test": "xo && nyc jasmine"
   },
   "author": "Dane Thurber <dane.thurber@gmail.com>",
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -50,5 +51,18 @@
...
       "spec"
     ]
+  },
+  "xo": {
+    "space": 2,
+    "rules": {
+      "space-before-function-paren": ["error", "never"]
+    },
+    "globals": [
+      "describe",
```
D34THWINGS: You can use:
```json
{
  "env": {
    "jasmine": true
  }
}
```
instead of adding every global one by one

mastilver: Damn, didn't know about this... Super useful 👍

D34THWINGS: FYI https://eslint.org/docs/user-guide/configuring#specifying-environments

shellscape:
I'm going to be taking this an ESLint route. Thanks for the work on this PR previously.

PR #75, opened 2017-08-23
Copy plugin integration

davidmpaz:
Hi @mastilver,
This is the follow up of danethurber/webpack-manifest-plugin#45. Unfortunately I will not have more time these days until next week for the project.

I need help here also with [these lines](https://github.com/davidmpaz/webpack-manifest-plugin/commit/d2cb47f2c8ebb522c7581ec888ee77537fb4e4c5#diff-f85c36a1cc8952fb5e6fc7200a89ceaaR31), in the function `existAssetInFiles`. From tests I was getting an edge case for files ending in: `.hot-update.json`. Currently the solution I found was that you see there. But honestly I dont know from where that comes and whether this is the right way to check if the asset was included already. Please could you take a look and improve from there?

Things we can have in mind: 

If there is something we could do from [copy-webpack-plugin](https://github.com/kevlened/copy-webpack-plugin) side to make easier that check, I can make a pull request to them. If you look at the [mock plugin](https://github.com/davidmpaz/webpack-manifest-plugin/commit/63cf231fd77e2aca692b47c4bf91ec7a403289b9) what it does is asimilar to what [copy-webpack-plugin](https://github.com/kevlened/copy-webpack-plugin) does. Maybe we can add another data there to use later in here to filter easily. Some flag that does not interfere with standard things and so on like:
```
compilation.assets['third.party.js'] = {
       copied: true, // this is what I mean as flag, though dunno if correct.
       size: function () {
         return compiledMock.length;
       },
       source: function () {
         return compiledMock;
       }
}
```
So [in here](https://github.com/davidmpaz/webpack-manifest-plugin/commit/9438c1f77f8d40769abb0acfcd6db046f0a5eed3#diff-f85c36a1cc8952fb5e6fc7200a89ceaaR94) we can do something like:
```
if (asset.copied) {
......
}
```

Sorry for the long post.
regards

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -28,6 +28,19 @@ ManifestPlugin.prototype.getFileType = function(str) {
   return ext;
 };
 
+ManifestPlugin.prototype.existAssetInFiles = function(asset, files) {
+
+  if (path.extname(path.basename(asset.name, '.json')) === '.hot-update') {
+    return true;
+  }
+
+  // name is not in files already
+  return undefined !== _.find(files, function(file) {
+    var chunkName = path.basename(file.name, path.extname(file.name));
+    return file.path === asset.name || asset.chunkNames.indexOf(chunkName) !== -1;
+  });
+};
+
 ManifestPlugin.prototype.apply = function(compiler) {
   var outputName = this.opts.fileName;
   var seed = this.opts.seed || this.opts.cache || {};
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -91,6 +91,17 @@ ManifestPlugin.prototype.apply = function(compiler) {
     files = stats.assets.reduce(function (files, asset) {
       var name = moduleAssets[asset.name];
       if (!name) {
+        if (!this.existAssetInFiles(asset, files)) {
+          files.push({
+            path: asset.name,
+            name: asset.name,
+            isInitial: false,
+            isChunk: false,
+            isAsset: true,
+            isModuleAsset: false
+          });
+        }
+
         return files;
       }
 
@@ -102,7 +113,7 @@ ManifestPlugin.prototype.apply = function(compiler) {
         isAsset: true,
         isModuleAsset: true
       });
-    }, files);
+    }.bind(this), files);
 
     // Append optional basepath onto all references.
     // This allows output path to be reflected in the manifest.
--- a/spec/helpers/copy-plugin-mock.js
+++ b/spec/helpers/copy-plugin-mock.js
@@ -1,0 +1,26 @@
+function FakeCopyWebpackPlugin() {
+};
+
+/**
+ * Mock plugins that modifies `compilation.assets`
+ *
+ * @param compiler
+ */
+FakeCopyWebpackPlugin.prototype.apply = function (compiler) {
+  compiler.plugin('emit', function (compilation, callback) {
+
+    var compiledMock = '// some compilation result\n';
+    compilation.assets['third.party.js'] = {
+      size: function () {
+        return compiledMock.length;
+      },
+      source: function () {
+        return compiledMock;
+      }
+    };
+
+    callback();
+  });
+};
+
+module.exports = FakeCopyWebpackPlugin;
--- a/spec/helpers/copy-plugin-mock.js
+++ b/spec/helpers/copy-plugin-mock.js
@@ -1,11 +1,6 @@
 function FakeCopyWebpackPlugin() {
 };
 
-/**
- * Mock plugins that modifies `compilation.assets`
- *
- * @param compiler
- */
 FakeCopyWebpackPlugin.prototype.apply = function (compiler) {
   compiler.plugin('emit', function (compilation, callback) {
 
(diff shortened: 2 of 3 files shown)


mastilver:
Great  thank you, I will have a look at it later

Sorry, if it was a bit painful due to all the changes :)

mattandrews:
This is a feature I need – I've tried using this branch in my `package.json` but I'm still not seeing copied files appear in my manifest. This is my config:

    plugins: [
            new CopyWebpackPlugin([{
                from: './images',
                to: './images',
                context: path.resolve(__dirname, 'assets')
            }]),
            new WebpackAssetsManifest({
                output: 'manifest.json',
                writeToDisk: true
            })
        ]

... `manifest.json` has my JavaScript files correctly listed, but not the images copied in `CopyWebpackPlugin`. Looking at these changes, it doesn't seem like I need to pass any additional options, but am I using it wrongly? I should expect to see the copied image files in my manifest file here, right?

mastilver:
@mattandrews Can you double you have this `"webpack-manifest-plugin": "davidmpaz/webpack-manifest-plugin#copy-plugin"` in your `package.json`

mattandrews:
Hi @mastilver – this is what I have:

    "webpack-manifest-plugin": "git://github.com/davidmpaz/webpack-manifest-plugin.git#copy-plugin"

When I browse the files it installs in `node_modules`, I can confirm it has the changes in this PR (eg. the `existAssetInFiles` function, etc).

mastilver:
I think you will have to check with @davidmpaz, I have zero experience with that plugin..., but it's an issue if it doesn't work.
Can you give us your config or better a repository to reproduce it

davidmpaz:
Would be nice to have some more data to look at. Just tested again the branch and I see no problem.

Can you provide requested info @mattandrews ?

mattandrews:
Thanks both – I've set up a simple repo here which shows an issue: https://github.com/mattandrews/webpack-manifest-copy-images-bug

A couple of things to note though:

- I was using the wrong plugin in my initial attempts (I was using [webpack-assets-manifest](https://github.com/webdeveric/webpack-assets-manifest) instead of this plugin – sorry, this was a sloppy mistake)

- Despite this, when using @davidmpaz's fork of the plugin, I don't get the output I expect in my test repo above. It's probable I'm not using Webpack correctly (I'm still pretty new to it) – I'm coming from a `gulp` workflow so trying to copy my files to an output directory, version them, then generate a manifest that contains their input/output pairings. It seems as though the `copy-webpack-plugin` (which I appreciate you have no experience of, @mastilver!) doesn't pass the pre-hashed filename properly through to this fork?

To be honest this issue may be down to my misunderstanding of Webpack (or just a bad workflow I'm trying to create?) rather than an initial bug with this fork – certainly my report above was wrong because I wasn't even using this plugin 😞).

mastilver:
@mattandrews I'm afraid this can't really be solved... unless you can have access to the `written` variable? https://github.com/kevlened/copy-webpack-plugin/blob/de4ff3231646548e9b524cde45cc1655cbb3373d/src/writeFile.js#L73

davidmpaz:
Hi @mattandrews 

don't worry too much about it, we all learn new stuff everyday :) Like @mastilver said is a bit complicate because of how copy-plugin works. From my experience, getting files copied with hashed file names for versioning was not working as expected. 

One thing I did tried was to hash files when copying everything just one level under directory to copy from. That worked fine. Still when trying to use with globs could not make it work, honestly did not look much more deeper later on.

Some threads you could read, maybe can help to find a solution regarding versioning copied files:
* copy-webpack-plugin#105
* copy-webpack-plugin#104

Regarding manifest plugin, just be sure to add the [copy-webpack-plugin](https://github.com/kevlened/copy-webpack-plugin) before the **manifest plugin** in order to get all copied files included in the webpack compilation.

If you think this issue is not relevant any more please feel free to close it :)

cheers

mastilver:
Thank you for your work @davidmpaz :)

bxm156:
This worked great for me. I needed it to copy django's static files into webpack to get them into my build process and manifest.json Much appreciated!

Levdbas:
Currently on the RC2. Does this also work with hashed filenames? Currently using the CopyWebPackPlugin like this:

`new CopyWebpackPlugin([{
      from: 'assets/images/',
      to: process.env.NODE_ENV === 'production' ? 'images/[name].[hash].[ext]' : 'images/[name].[ext]'
    }]`

Expected output in manifest:

`"images/file.jpg": "images/file.bf95ed4f733b97f45d2af4456872c2c3.jpg",`

what gets outputted in the manifest: 

`"images/file.bf95ed4f733b97f45d2af4456872c2c3.jpg": "images/file.bf95ed4f733b97f45d2af4456872c2c3.jpg",`

Am I missing something?

mattandrews:
@Levdbas I ended up not using this plugin and used a gulp task to version all my assets at build time (eg. for deployments, not day-to-day development). It simplified things as local development doesn't need versioning, and therefore I didn't need to tell Webpack about a bunch of unrelated files (image assets etc).

Levdbas:
Thanks @mattandrews for your approach. Will consider someting like that as well if this really cant be solved.

@mastilver could the guys from copywebpack or any other plugin implement this new hook in such a way that no matter if the filename is hashed or not, it still gets a properly output in the manifest? In that case I will head over to the copywebpack repo to ask :)

deguif:
See my https://github.com/webpack-contrib/copy-webpack-plugin/issues/104#issuecomment-370174211 for a working solution using `copy-webpack-plugin`.

PR #78, opened 2017-09-01
fix: make sure paths are consistent from os to os

### LINKED ISSUE #60: Option to map entries or normalize paths on both Windows and OSX/Linux
Hi,

The keys of the manifest on Windows are: `"path\\file.ext"`

Whereas on Linux/OSX it is: `"path/file.ext"`

Making it difficult to reference a resource by its key in the manifest.

Can we have a callback function to transform the manifest elements prior to insertion or have both paths use slashes (`/`)?

mastilver:
closes #60

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -110,6 +110,12 @@ ManifestPlugin.prototype.apply = function(compiler) {
       }.bind(this));
     }
 
+    files = files.map(file => {
+      file.name = file.name.replace(/\\/g, '/');
+      file.path = file.path.replace(/\\/g, '/');
+      return file;
+    });
+
     if (this.opts.filter) {
       files = files.filter(this.opts.filter);
     }
--- a/spec/plugin.spec.js
+++ b/spec/plugin.spec.js
@@ -398,6 +398,24 @@ describe('ManifestPlugin', function() {
         done();
       });
     });
+
+    it('should output unix paths', function(done) {
+      webpackCompile({
+        context: __dirname,
+        entry: {
+          'dir\\main': './fixtures/file.js',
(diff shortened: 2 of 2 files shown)


lib/plugin.js
```diff
@@ -110,6 +110,12 @@ ManifestPlugin.prototype.apply = function(compiler) {
       }.bind(this));
     }
 
+    files = files.map(file => {
+      file.name = file.name.replace('\\', '/');
+      file.path = file.path.replace('\\', '/');
```
a-x-: `'\\'` is equivalent `/\//`
but `/\//g` expected

a-x-:
```js
'sdfds\\sdfsdf\\'.replace('\\', '/') // -> "sdfds/sdfsdf\"
'sdfds\\sdfsdf\\'.replace(/\\/, '/') // -> "sdfds/sdfsdf\"
'sdfds\\sdfsdf\\'.replace(/\\/g, '/') // -> "sdfds/sdfsdf/"
```

spec/plugin.spec.js
```diff
@@ -398,6 +398,22 @@ describe('ManifestPlugin', function() {
         done();
       });
     });
+
+    it('should output unix paths', function(done) {
+      webpackCompile({
+        context: __dirname,
+        entry: {
+          'dir\\main': './fixtures/file.js'
```
a-x-: add `some\\dir\\main also`

lib/plugin.js
mastilver: Good catch :+1:

mastilver:
@a-x- fixed the issue you spotted :)

PR #90, opened 2017-09-16
feat: add generate option

mastilver:
BREAKING CHANGE: reduce option replaced by generate option

closes #89 

@SeeThruHead Does that look goo to you?

PR #92, opened 2017-09-20
docs: remove leftover reduce setting

joscha:
Fixed some leftovers from #90. It's also quite confusing currently to have the docs for the new `generate` option, but it not being released yet. Any `opts.generate` is just silently ignored at the moment, I think either a release or backwards compatibility would be good.


mastilver:
Thank you @joscha 

I will be releasing `v2` by the end of the week, I know I could have done a better job for crafting this release, but that's the way it is now ;)

See https://github.com/danethurber/webpack-manifest-plugin/blob/1.x/README.md for current docs

PR #93, opened 2017-09-21
adds topological sort example

SeeThruHead:
add topological sort to new examples folder
adds examples readme entry with link to specific example
gitignore example output files

prior work: #79

### DIFF
--- a/examples/topological-sort/webpack.config.js
+++ b/examples/topological-sort/webpack.config.js
@@ -1,0 +1,53 @@
+var plugin = require('../../index.js');
+var webpack = require('webpack');
+var _ = require('lodash');
+var toposort = require('toposort');
+
+function generate(seed, files) {
+  var nodeMap = {};
+
+  files.forEach(function (file) {
+    nodeMap[file.chunk.id] = file;
+  });
+
+  var edges = [];
+
+  files.forEach(function (file) {
+    if (file.chunk.parents) {
+      file.chunk.parents.forEach(function (parent) {
+        var parentFile = _.isObject(parent) ? nodeMap[parent.id] : nodeMap[parent];
+        if (parentFile) {
+          edges.push([parentFile, file]);
+        }
+      });
+    }
+  });
+
+  const sortedFiles = toposort.array(files, edges);
+
+  return sortedFiles.reduce(function(manifest, file) {
+    return manifest.concat({ name: file.name, path: file.path });
+  }, seed);
+}
+
+
+module.exports = {
+  context: __dirname,
+  entry: {
+    main: './main.js',
+    vendor: './util.js'
+  },
+  output: {
+    filename: '[name].out.js'
+  },
+  plugins: [
+    new plugin({
+      seed: [],
+      generate: generate
+    }),
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'common',
+      filename: 'common.out.js'
+    })
+  ]
+};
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -41,7 +41,8 @@
   "dependencies": {
     "fs-extra": "^0.30.0",
     "lodash": ">=3.5 <5",
-    "mutexify": "1.0.1"
+    "mutexify": "1.0.1",
+    "toposort": "^1.0.3"
```
mastilver: Can you move that to `devDependencies`

examples/topological-sort/webpack.config.js
```diff
@@ -0,0 +1,53 @@
...
+
+
+module.exports = {
+  context: __dirname,
+  entry: {
+    main: './main.js',
+    vendor: './util.js'
+  },
+  output: {
+    filename: '[name].out.js'
```
mastilver: can you set `output.path` to `path.join(__dirname, 'build')` and `filename` to `[name].js`

I've got plan for the future to use those in tests, that will make it easier

examples/topological-sort/webpack.config.js
```diff
@@ -0,0 +1,53 @@
+var plugin = require('../../index.js');
+var webpack = require('webpack');
+var _ = require('lodash');
+var toposort = require('toposort');
+
+function generate(seed, files) {
+  var nodeMap = {};
+
+  files.forEach(function (file) {
+    nodeMap[file.chunk.id] = file;
```
joscha: I think this needs an `isChunk` guard.

examples/topological-sort/webpack.config.js
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  var edges = [];
+
+  files.forEach(function (file) {
+    if (file.chunk.parents) {
```
joscha: same here

SeeThruHead:
i can take a look at this tmrw. what better way to spend my airport wait :D

examples/topological-sort/webpack.config.js
SeeThruHead: so i'm not actually all that familiar with webpack, what kind of dependency tree would give me files without chunks? I supposed the example should handle that case? or is it sufficient to handle the actual case that the example files dep tree creates (all files are chunks?)

mastilver:
Hey @SeeThruHead :)

Let me know how you are getting on with that. Do you want me to take it over (I will have more time in a couple of weeks)?

SeeThruHead:
hey @mastilver I made a comment above asking for some guidance with how to test an isChunk guard. ie what kind of dep tree would give me files without chunks.

examples/topological-sort/webpack.config.js
mastilver: Oops sorry, missed that comment.

We include emitted files (for example from `copy-webpack-plugin`), basically it's every non-js files. I think those better be at the end (but really I don't care, up to you!)

caub:
it's a bit brutal, but I've done it here https://github.com/caub/webpack-manifest-plugin/blob/m2/lib/plugin.js (the logic actually mostly copied from HtmlWebpackPlugin)

I think the manifest file deserves to be sorted by default

mastilver:
html-webpack-plugin is getting rid of toposort, in flavours of `entrypoints`, we should do the same for this plugin I believe

shellscape:
@SeeThruHead Thanks for opening this PR a while back, and I'm sorry that it didn't get attention sooner. We've just landed a major refactor in #222, and part of that is only supporting webpack v4.44.0+. Please let me know if you're interested in reopening this PR and we'll move forward.

PR #95, opened 2017-09-25
fix: default sort option

joscha:


mastilver:
Yep, missed that. Thank you!

Issue #100, opened 2017-10-30
having multiple instances of this plugin seems to prevent webpack from emitting any files other than the manifests


BernsteinA:
2.0.0-rc.1

```
    new ManifestPlugin({
        fileName: 'manifest-android.json',
        filter: (file)=>!file.path.match(/\.mp3$|\.map$/),
    }),
    new ManifestPlugin({
        fileName: 'manifest-ios.json',
        filter: (file)=>!file.path.match(/\.ogg$|\.map$/),
    })
```

with only one ManifestPlugin everything works. with two, no assets are emitted by webpack, except the two manifest jsons.

mastilver:
Interesting... I wonder if it's due to: https://github.com/danethurber/webpack-manifest-plugin/blob/409b0fb64234ed9fa9adeeba31850a4d4feeca7c/lib/plugin.js#L173-L178

Could you send me a failing PR, please

BernsteinA:
Not sure what you mean by a failing PR. Add a test to https://github.com/danethurber/webpack-manifest-plugin/blob/master/spec/plugin.integration.spec.js ?

mastilver:
yes, or rather: spec/plugin.spec.js

mastilver:
So I don't think I'm going to fix that

I believe it's bad practice to have multiple instances of the same plugins (I could be wrong)

But I think we can work around that: `manifest-webpack-plugin` could accept an array of options

```js
 new ManifestPlugin([
    {
        fileName: 'manifest-android.json',
        filter: (file)=>!file.path.match(/\.mp3$|\.map$/),
    },
    {
        fileName: 'manifest-ios.json',
        filter: (file)=>!file.path.match(/\.ogg$|\.map$/),
    }
])
```

What do you think?

BernsteinA:
I haven't seen anything that suggests it's bad practise to have multiple instances of the same plugin.
https://github.com/jantimon/html-webpack-plugin#generating-multiple-html-files
https://webpack.js.org/plugins/commons-chunk-plugin/#combining-implicit-common-vendor-chunks-and-manifest-file

mastilver:
Thank you for the `CommonsChunkPlugin` example 👍 

I'm one of the maintainer of `html-webpack-plugin` and issues with multiple instances are just a pain to deal with ;)

Anyway I will try to see if I can fix it (by allowing multiple instances), feel free to have a look on your side as well :)

a-x-:
[...]

PR #101, opened 2017-10-31
added (failing) test for issue #100

BernsteinA:
#100 💯 
webpack never finishes. jasmine times out.

### DIFF
--- a/spec/plugin.integration.spec.js
+++ b/spec/plugin.integration.spec.js
@@ -65,6 +65,40 @@ describe('ManifestPlugin using real fs', function() {
       });
     });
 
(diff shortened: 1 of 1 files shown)


mastilver:
👍  great, I will need to have a look at that

Thank you for trying the rc, I was about to release it but it looks like it's not ready yet ;)

mastilver:
So yeah... removing https://github.com/danethurber/webpack-manifest-plugin/blob/409b0fb64234ed9fa9adeeba31850a4d4feeca7c/lib/plugin.js#L173-L181 fix the issue... :/

mastilver:
moved this commit on #107

PR #106, opened 2017-11-21
changed default filename from 'manifest.json' to 'manifest.webmanifest'

mojoaxel:
Thanks for this plugin! 👍 

This pull requests changes the default filename from `manifest.json` to `manifest.webmanifest` because this is (now) the recommended filename and extension by the W3C:
* https://www.w3.org/TR/appmanifest/#media-type-registration
* https://www.w3.org/TR/appmanifest/#using-a-link-element-to-link-to-a-manifest

Here is the discussion:
* https://github.com/w3c/manifest/issues/346
* https://github.com/w3c/manifest/issues/381

Some browser vendors also already adopted there docs:
* https://developer.mozilla.org/en-US/docs/Web/Manifest

_I would consider this a major change so maybe this is something for the major 2.0 release?_

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ module.exports = {
 };
 ```
 
-This will generate a `manifest.json` file in your root output directory with a mapping of all source file names to their corresponding output file, for example:
+This will generate a `manifest.webmanifest` file in your root output directory with a mapping of all source file names to their corresponding output file, for example:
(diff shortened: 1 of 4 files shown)


mojoaxel:
Sorry my mistake. Wrong manifest.json ;-)

PR #107, opened 2017-11-22
fix: use multiple instances at the same time 

### LINKED ISSUE #100: having multiple instances of this plugin seems to prevent webpack from emitting any files other than the manifests

mastilver:
closes #100

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -1,9 +1,8 @@
 var path = require('path');
 var fse = require('fs-extra');
 var _ = require('lodash');
-var mutexify = require('mutexify');
 
-var lock = mutexify();
+var manifestMap = {};
 
 function ManifestPlugin(opts) {
   this.opts = _.assign({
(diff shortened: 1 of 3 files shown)


deguif:
Regarding #100 discussion that means that's the multi instance solution that was chosen.

mastilver:
@deguif Yeah, I really wasn't sure what was the right solution..., I think this is the most natural for the user, I hope it doesn't introduce any bugs...

deguif:
@mastilver Yes, I totally agree. I find it better too. But from the discussion I was ready to play with an array of config.

deguif:
Is there any reason why it doesn't get merged?

deguif:
Any news about this PR?

PR #112, opened 2018-01-04
Use individual lodash packages

woodyrew:
Only two packages from lodash are used. These have been added appropriately.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -1,12 +1,12 @@
 var path = require('path');
 var fse = require('fs-extra');
-var _ = require('lodash');
+var assign = require('lodash.assign');
 var mutexify = require('mutexify');
 
 var lock = mutexify();
 
 function ManifestPlugin(opts) {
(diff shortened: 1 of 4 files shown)


spoike:
Is `assign` from lodash necessary?

[`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) has been in Node since [version 4](http://node.green/#ES2015-built-in-extensions-Object-static-methods-Object-assign) and this project's `package.json` specifies supporting Node `>= 4`.

spoike:
This PR will most likely be in conflict with changes in #58 (suggestion to remove lodash completely).

woodyrew:
#58 and #166 have the same result, neither have been merged and there are conflicts now.

The full [Lodash package is still in master](https://github.com/danethurber/webpack-manifest-plugin/blob/master/package.json#L45) `¯\_(ツ)_/¯`

Issue #110, opened 2017-12-08
Keys in manifest not actually source file names


jacobwindsor:
In the docs, it states "This will generate a manifest.json file in your root output directory with a mapping of all source file names to their corresponding output file..."

However, in actuality, the key is generated from the entry name. E.g.

```js
// webpack.config.js
...
entry: {
        'myentry': './path/to/my/entry/index.js'
    },
...
```

gives:
```
// manifest.json
{
  "myentry.js": "/dist/myentry.bundle.js",
}
```

I think this is actually better behaviour then simply using the source file name. However, the only inconsistency is the addition of the `.js` suffix in manifest.json. There should be an option to use the entry name (defined in webpack.config.js) or the source file name.

a-x-:
It makes sense for me

Keys are filenames with an assumption that webpack `output.filename` has a default value `[name].js`

What about `entryNameKeys: Bool` option?

a-x-:
possible code place: https://github.com/danethurber/webpack-manifest-plugin/blob/master/lib/plugin.js#L54

a-x-:
Alternatively we can create optional function `formatKey` with some arguments like `{ entryName, ext, chunk, files }` with webpack plugin context (`call(this, ...)`)

jacobwindsor:
The `entryNameKeys ` option seems more appropriate to me. And with what the docs say, should be true by default. If the addition of the js extension was by default (as is now) the docs should be reworded to clarify that. The entry name keys in manifest aren't actually the source file names, they are the entry names with a js extension. While a lot of people would have the entry name to be the same as the source file name, I think assuming that is dangerous.

Happy to make a PR once a decision has been made.

a-x-:
I agree,
may be we will create major release some time with entry name keys by default

I think it's possible to add an some option for now

will see what @mastilver position

jacobwindsor:
@a-x- @mastilver Since there was no activity on this for a while, I went ahead and created a PR hoping everyone agrees. I need this functionality for my own use case so just trying to speed things up.

Hope it's okay!

shellscape:
Fixed in a998c4b

PR #113, opened 2018-01-08
Add an entryNameKeys option

jacobwindsor:
Addresses #110

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,14 @@ Default: `false`
 
 If set to `true` will emit to build folder and memory in combination with `webpack-dev-server`
 
+### `options.entryNameKeys`
+
+Type: `Boolean`<br>
+Default: `false`
+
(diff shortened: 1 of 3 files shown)


jacobwindsor:
In the previous commit I have simply excluded sourcemaps from having their extension excluded. I don't think this is the best way to go. Instead, I would prefer something like: 

```javascript
{
   entryName: {
    source: 'main.js'
    sourceMap: 'main.js.map'
  }
}
```

However, that's a fairly large change and might warrant another PR entirely.

shellscape:
@jacobwindsor Thanks for opening this PR a while back, and I'm sorry that it didn't get attention sooner. We've landed a major refactor in #222 that has resulted in some significant conflicts. Would you be interested in reworking your PR on top of the new codebase?

shellscape:
Closed by a998c4bbd205d0bbf3c19b567ce259894c1502d0

Issue #122, opened 2018-02-20
Allow content serialization to be modified


tylerbwilliams:
Currently, the content returned from the `generate` function is then passed to `JSON.stringify`.  We don't get the chance to use a different serialization model.

My use case is that I want to create a `manifest.properties` file instead of a `manifest.json` file.  I can write my own `generate` function to create the output as a valid properties file, but then `JSON.stringify` wraps it in quotes and escapes all my newlines.

Can we have a `serialize` option that defaults to `JSON.stringify`, for those who want their own strategy for this?

mastilver:
Can you send a PR please?

mastilver:
done in https://github.com/danethurber/webpack-manifest-plugin/pull/123

PR #123, opened 2018-02-21
add outputFormatter option

Ficik:
Support different manifest output formats.

In my case I need to generate manifest as yaml to load it as symfony config. With this I can replace JSON.stringify with YAML.stringify in my webpack configuration.


mastilver:
Thank you for that! :)

I kind of prefer `serialize` options from https://github.com/danethurber/webpack-manifest-plugin/issues/122

Ficik:
Sure, serialize sounds better. Renamed to serialize.

mastilver:
Thank you @Ficik for your work :+1:

mastilver:
published on `v2`

PR #128, opened 2018-02-26
Improve webpack 4 support (afterEmit issue)

andriijas:


mastilver:
Thank you, so much :star: 

The failing test is there: https://github.com/danethurber/webpack-manifest-plugin/blob/master/spec/plugin.integration.spec.js#L74

radum:
@andriijas You should remove Node 4 from Travis also.

mastilver:
Yeah, good point! node@4 will be deprecated in April, so it's pretty safe to do it now

andriijas:
Tests are running....  💥

andriijas:
Call for help. 😱😱😱

I went through 💍 and 🔥 to get the tests working properly on all versions of webpack and extract text plugin and current status is that tests are running/failing randomly. 

Check https://travis-ci.org/danethurber/webpack-manifest-plugin/builds/347232114

```
    TypeError: used is not a function
```

I have no clue what that means, i get it locally sometimes as well.

mastilver:
Don't worry about `SyntaxError: Unexpected end of JSON input` it's just a racing condition issue (which we will be able to get rid once stop support for webpack < 3)

`TypeError: used is not a function` is a known issue https://github.com/mafintosh/mutexify/issues/7, just leave the current version for `mutexify`. I'm getting rid of it anyway in #107 so you are good :)

andriijas:
@mastilver okay cool. then the bigger question is whatever compiler.hooks.webpackManifestPluginAfterEmit  should use a sync or async API

check https://github.com/danethurber/webpack-manifest-plugin/pull/128/commits/65f826da547706f32165197f3e36d59f3c4c1b9e

I will squash the commits and force push when you have decided, dont want that ugly console.log in the history.

mastilver:
I think we want it sync for the future (So we won't need the lock system) but I could be wrong

I'm sorry, I just merged something and there is some merge conflicts now... :/

andriijas:
@mastilver rebased.

lib/plugin.js
```diff
@@ -214,15 +211,20 @@ ManifestPlugin.prototype.apply = function(compiler) {
...
+
+    compiler.hooks.webpackManifestPluginAfterEmit = new SyncWaterfallHook(['manifest']);
+
+    compiler.hooks.compilation.tap('ManifestPlugin', compilation => {
       compilation.hooks.moduleAsset.tap('ManifestPlugin', moduleAsset);
     });
     compiler.hooks.emit.tap('ManifestPlugin', emit);
   } else {
-    compiler.plugin('compilation', function (compilation) {
+    compiler.plugin('compilation', compilation => {
```
aulisius: Shouldn't this also be `compiler.hooks.compilation.tap('ManifestPlugin', compilation)` ?

andriijas: Nah thats the legacy for webpack <=3

aulisius: Ah, ok.

mastilver:
We need to figure out where the memory leak come from, I did some tweaking but I was unsuccessful... :/

andriijas:
extract-text-webpack-plugin  is now in maintainance mode for webpack 3.

Maybe we should add https://github.com/webpack-contrib/mini-css-extract-plugin  for webpack 4 tests.

it will require some work though

mastilver:
Oh, so `extract-text-webpack-plugin` won't ever be released for webpack 4?

I think we should just ignore that test for now... release v2 then release v3 with just support for webpack@4 (the failing test is something that is happening very rarely)

Maybe check if we are running webpack@4, then return from the test

What do you think?

andriijas:
Sounds good👍

rossta:
@mastilver Looks like webpack 4 support is being addressed on webpack-contrib/extract-text-webpack-plugin#685

montogeek:
Please take a look at https://github.com/webpack-contrib/mini-css-extract-plugin, it was build for webpack 4 as a replacement for `extract-text-webpack-plugin`, please note that as its name says, it is intended for CSS extraction, no any text like `extract-text-webpack-plugin`.

andriijas:
@rossta nope thats obsolete PR. check webpack-contrib/extract-text-webpack-plugin#749

@mastilver updated PR with some branching/if statements to detect webpack version in tests. Added test with MiniCSSExtractPlugin for webpack >=4

didnt have time for this but it had to be done 😂

PR #132, opened 2018-03-15
allow name to include its own ext

robbtraister:
Don't append a file extension if the name already includes one.


radum:
@robbtraister I was a bit curious what is the use case for this? Can you share an example when this is needed?

shellscape:
Thanks for opening this PR a while back, and I'm sorry that it didn't get attention sooner. Since the last comment was over two years ago, we're going to close this as abandoned.

PR #142, opened 2018-04-12
fix: always run hooks last for `webpack@4`

### LINKED ISSUE #141: Use `stage: Infinity` for `webpack@4`
A lot of plugins use `emit` event to add own assets, manifest in 99% required only when all files were added to `compilation.assets`. Tapable have `stage` option for avoid this situation. https://github.com/webpack/tapable/blob/6b91ca9e94cf032420a57a2d0b82b627462b5c1c/lib/Hook.js#L130. 

Just change from:
```js
compiler.hooks.emit.tap('ManifestPlugin', emit);
```
To:
```js
compiler.hooks.emit.tap({ name: 'ManifestPlugin', stage: Infinity}, emit);
```

It is allow to avoid situation when other plugin use `emit` event and add new assets. I agree what using hook is better, but sometimes it is impossible, because you don't have access to plugin repo.

evilebottnawi:
fixes #141

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -203,15 +203,19 @@ ManifestPlugin.prototype.apply = function(compiler) {
 
   if (compiler.hooks) {
     const SyncWaterfallHook = require('tapable').SyncWaterfallHook;
+    const pluginOptions = {
+      name: 'ManifestPlugin',
+      stage: Infinity
+    };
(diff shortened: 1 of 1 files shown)


evilebottnawi:
Will be great after merge release patch version :+1:

PR #146, opened 2018-04-24
move emitCountMap to support multiple webpack configs

philiplehmann:
only one manifest file was written for multiple configs:
```
const Webpack = require('webpack')
const WebpackDevServer = require('webpack-dev-server')
const config1 = {...}
const config2 = {...}
const config3 = {...}
const config4 = {...}

const compiler = Webpack([config1, config2, config3, config4])
const server = new WebpackDevServer(compiler, devServerConfig)
server.listen(PORT, HOST)
```

moved the emitCountMap into the apply function. before it was global and shared between the plugin instances.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -2,8 +2,6 @@ var path = require('path');
 var fse = require('fs-extra');
 var _ = require('lodash');
 
-const emitCountMap = new Map();
-
 function ManifestPlugin(opts) {
   this.opts = _.assign({
     publicPath: null,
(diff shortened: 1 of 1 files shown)


mastilver:
It should already support multiple configs, see: https://github.com/danethurber/webpack-manifest-plugin/blob/master/spec/plugin.integration.spec.js#L24

philiplehmann:
what i wanted is to get a manifest file for each configuration. which it did in a earlier version. maybe an option would be better. i see now, that this was changed intentional.

mastilver:
I fail to understand what you want... the example you provided above should work. Can you send us a failing test so we can look into it

philiplehmann:
sounds good

will do the changes

PR #154, opened 2018-06-19
remove .bind and map -> forEach

caub:


shellscape:
Thanks for opening this PR a while back, and I'm sorry that it didn't get attention sooner. We've landed a major refactor in #222 that has resulted in some significant changes, and some which has rendered your changes in this PR unneeded. Thanks again for contributing and I hope you'll consider doing so again in the future.

PR #156, opened 2018-06-25
Sort alphabetically

Tymek:


shellscape:
Thanks for opening this PR a while back, and I'm sorry that it didn't get attention sooner. This is exactly the kind of customization that the `sort` option is supposed to empower users to be able to perform. Unfortunately we won't be adding it to the our feature set.

PR #174, opened 2018-10-23
Fix high memory usage (#173)

### LINKED ISSUE #173: Very high memory usage (~10GB for ~3000 files)
When the compilation uses A LOT of files, the `ManifestPlugin` uses A LOT of memory.

As an example, our compilation has ~2800 files (only assets: images, videos, documents), the `ManifestPlugin` used between 8 & 10GB of RAM to complete. _This is way too much!_

I've tracked the issue to be at line:
https://github.com/danethurber/webpack-manifest-plugin/blob/bcca8907df46ad6b57199d34d7be876f95170a52/lib/plugin.js#L58
And more specifically, the `toJson()` call.

This method is used to get easy-to-use object with information about the webpack compilation.
The plugin only uses the `assets` field of the generated object, here:
https://github.com/danethurber/webpack-manifest-plugin/blob/bcca8907df46ad6b57199d34d7be876f95170a52/lib/plugin.js#L88

The very high memory usage happens while generating the fields related to `modules`, an easy fix is to disable this part:

```diff
-var stats = compilation.getStats().toJson();
+var stats = compilation.getStats().toJson({
+  modules: false,
+});
```

It'll still generate other data that the plugin doesn't use.
Another fix is to disable everything we don't need:
```diff
-var stats = compilation.getStats().toJson();
+var stats = compilation.getStats().toJson({
+  version:     false,
+  hash:        false,
+  timings:     false,
+  builtAt:     false,
+  env:         false,
+  publicPath:  false,
+  outputPath:  false,
+  entrypoints: false,
+  chunks:      false,
+  chunkGroups: false,
+  children:    false,
+  modules:     false, // the memory hungry part is disabled here
+
+  assets: true, // only get assets information
+});
```

These two diffs resolves the problem, and the compilation uses normal amount of memory.

Let me know what you think about that

bew:
Fixes #173 

The `toJson()` call can use a lot of memory by generating a lot of stats information about modules.
But we don't use this information for this plugin, so we can only enable the generation of assets related data used for this plugin.

Read #173 for more information

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -55,7 +55,13 @@ ManifestPlugin.prototype.apply = function(compiler) {
     var seed = this.opts.seed || {};
 
     var publicPath = this.opts.publicPath != null ? this.opts.publicPath : compilation.options.output.publicPath;
-    var stats = compilation.getStats().toJson();
+    var stats = compilation.getStats().toJson({
+        // Disable data generation of everything we don't use
+        all: false,
+
+        // Enable generation of `assets` data
+        assets: true,
+    });
 
     var files = compilation.chunks.reduce(function(files, chunk) {
       return chunk.files.reduce(function (files, path) {


bew:
Based on https://github.com/webdeveric/webpack-assets-manifest/pull/47 I simplified the implementation to be future-proof (effectively disabling everything by default instead of manually listing every field we don't want one by one).

Kenneth-KT:
Not sure if the author still want to maintain this library.

If not, there is another plugin [webpack-assets-manifest](https://github.com/webdeveric/webpack-assets-manifest) that do exactly the same thing and they have this problem fixed in [this commit](https://github.com/webdeveric/webpack-assets-manifest/commit/051243ec8963be69ab8565b5998032cc3a020a7a) already, but they haven't released it yet.

Run `yarn add --dev https://github.com/webdeveric/webpack-assets-manifest.git\#051243e` to add it to your repo.

mastilver:
Hi @bew

Could you fix the tests please?

PR #182, opened 2018-12-16
Fix missing array in Readme

stephtr:
The readme currently has (at least two) errors:
`sort` functions should have two arguments
the second argument for `generate` should be an array


shellscape:
Thanks for opening this PR a while back, and I'm sorry that it didn't get attention sooner. Due to a major refactor in #222, we've reimplemented your fix in this commit: 7bafd43ee154f917211593ed2b034bd0ccaacdf2

PR #184, opened 2019-02-05
Update README to note abs output location option

damassi:
Updates README to note that fileName can accept a path to a location.


mastilver:
Thank you, I was just suggesting that in #187

PR #187, opened 2019-05-03
Add outputPath to options

darshanrajani:


mastilver:
You can pass an abstract path to https://github.com/danethurber/webpack-manifest-plugin#optionsfilename

If you can send a PR that clarify it in the readme that would be awesome!

PR #191, opened 2019-09-13
Store custom hook in WeakMap for Webpack 5 compatability

### LINKED ISSUE #186: Support webpack@5
```
TypeError: Cannot read property 'call' of undefined
    at ManifestPlugin.<anonymous> (/.../node_modules/webpack-manifest-plugin/lib/plugin.js:191:53)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/.../node_modules/tapable/lib/HookCodeFactory.js:32:10), <anonymous>:9:1)
    at AsyncSeriesHook.lazyCompileHook (/.../node_modules/tapable/lib/Hook.js:154:20)
    at Compiler.emitAssets (/.../node_modules/webpack/lib/Compiler.js:448:19)
    at process.nextTick (/.../node_modules/webpack/lib/Compiler.js:249:10)
    at processTicksAndRejections (internal/process/next_tick.js:74:9)
(node:28406) DeprecationWarning: Compilation.chunks was changed from Array to Set (using Array method 'reduce' is deprecated)
```

ian-craig:
From Webpack 5, it's no longer possible to just assign new custom hooks onto the `compiler.hooks` object. The recommendation is to store these in a `WeakMap` keyed off the compiler instance.

https://github.com/webpack/changelog-v5/blob/master/README.md#hook-object-frozen

Running plugin v2.0.4 in Webpack v5.0.0-alpha.26 gives the following error:
```
TypeError: Cannot read property 'call' of undefined
    at ManifestPlugin.<anonymous> (...\node_modules\webpack-manifest-plugin\lib\plugin.js:191:53)
    at Hook.eval [as callAsync] (eval at create (...\node_modules\webpack\node_modules\tapable\lib\HookCodeFa
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (...\node_modules\webpack\node_modules\tapable\lib\Hook.js:18
    at Compiler.emitAssets (...\node_modules\webpack\lib\Compiler.js:507:19)
    at process.nextTick (...\node_modules\webpack\lib\Compiler.js:302:10)
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

This change just moves the storage of `webpackManifestPluginAfterEmit` into a WeakMap.

Tested in Webpack v4.35.3 and v5.0.0-alpha.26.

Fixes #186

**Update:**
I did a second pass over this and created a static method to expose custom hooks to other plugins, following the recommended approach and a similar implementation to [fork-ts-checker-webpack-plugin](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/166/files).

I'd expect other plugins to use hooks as follows:
```js
ManifestPlugin.getCompilerHooks(compiler).afterEmit.tap('ManifestPlugin', (manifest) => {
   //...
});
```

This new behavior is supported for Webpack 4+, but I've also preserved `compiler.hooks.webpackManifestPluginAfterEmit` in Webpack 4


mastilver:
Hi @ian-craig 

Could you update .travis.yml to test webpack@5 ?

spec/plugin.spec.js
```diff
@@ -615,7 +620,7 @@ describe('ManifestPlugin', function() {
         }
       }, function(manifest){
         expect(manifest).toEqual({
-          'javascripts/main.js': 'javascripts/main.js'
+          [`javascripts${path.sep}main.js`]: 'javascripts/main.js'
```
mastilver: note to self: Is this the expected behaviours?

spec/plugin.spec.js
```diff
@@ -407,9 +411,7 @@ describe('ManifestPlugin', function() {
           'main.js': 'main.js'
         });
 
-        expect(JSON.parse(stats.compilation.assets['manifest.json'].source())).toEqual({
-          'main.js': 'main.js'
-        });
+        expect(emittedAsset(stats.compilation, 'manifest.json')).toEqual(true)
```
mastilver: Could you check it's content?

spec/plugin.spec.js
mastilver: yep, it should stay 'javascripts/main.js', are you having issues on windows?

It should be covered by: https://github.com/danethurber/webpack-manifest-plugin/blob/421fa29716cdb26251c10e90c6d32ecc56b6a6c8/lib/plugin.js#L152-L156

spec/plugin.spec.js
ian-craig: It looks like Webpack 5 no longer puts the whole file in this object: https://github.com/webpack/changelog-v5/blob/master/README.md#sizeonlysource-after-emit

I could make this test specific to Webpack <= 4 if you want to still verify the content is available here for older versions?

spec/plugin.spec.js
ian-craig: The output in Windows is
```
{
    'javascripts\main.js': 'javascripts/main.js'
}
```

I found it odd that the value is changed byt not the key. If this isn't the expected behavoir, then I guess it's an unrelated bug. I'll revert this test change.

spec/plugin.spec.js
mastilver: NP, I'm planning on updating the tests once v3 is out to rely less on webpack internal

spec/plugin.spec.js
mastilver: I tried overwriting `path.sep` to simulate a windows environment, and the test pass

Could you double check it on your side?

ian-craig: Looks like the `opts.map` provided by the UT itself was changing the file name and using the OS path sep. I updated `plugin.js` to standardize paths again after opts.map is run as I figure you want the paths standardized both before `opts.filter` and after `opts.map`

gaurav5430:
Does this change the syntax for usage of the webpackManifestPluginAfterEmit in the consumers as well ?

We have been using this syntax in our own plugin to tap the manifest generation:

`compiler.hooks.webpackManifestPluginAfterEmit.tap` with version 2.2.0

do we need to change this ?

PR #194, opened 2019-09-24
Fix `entrypoints` compatibility for older Node versions

### LINKED ISSUE #193: SyntaxError: Unexpected token ...
Hi, I just updated to version 2.1.0 and I'm getting this error when building:
`SyntaxError: Unexpected token ...`

My config:
```
        plugins: [
          new ManifestPlugin({
            fileName: '../../../.js-names.json',
            filter: (file) => !file.path.match(/\.map$|\d-chunk/)
          })
        ],
```

Guessing it has problems to handle es6 features. Any idea how I could  fix this?

samuelmeuli:
I've introduced changes in #192 that aren't compatible with some older versions of Node.js. This PR fixes that. It can be reverted once the required Node.js version is increased.

Fixes #193

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -1,23 +1,9 @@
+var entries = require('object.entries');
 var path = require('path');
 var fse = require('fs-extra');
 var _ = require('lodash');
 
 const emitCountMap = new Map();
-
-// Polyfill for `Object.entries`
-// Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
-// TODO: Remove once support for older versions of Node.js is dropped
-if (!Object.entries) {
-  Object.entries = function( obj ){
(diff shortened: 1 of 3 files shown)


mastilver:
hey Samuel

Going to close this as we no longer support older node versions (see #193 for more info)

lib/plugin.js
```diff
@@ -4,6 +4,21 @@ var _ = require('lodash');
 
 const emitCountMap = new Map();
 
+// Polyfill for `Object.entries`
+// Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
+// TODO: Remove once support for older versions of Node.js is dropped
+if (!Object.entries) {
```
mastilver: Could you use https://www.npmjs.com/package/object.entries instead?

mastilver:
@samuelmeuli Could you install `object.entries`

PR #197, opened 2019-09-25
v3

mastilver:
closes #195
fixes #167

PR #200, opened 2019-10-10
Webpack 5: Add "ids" to the retrieved JSON stats

Lyrkan:
This PR adds the `ids` objects to the stats retrieved by `stats.toJson()`.

It is needed by the latest alpha of Webpack 5 (`v5.0.0-alpha.31`) in order to retrieve `asset.chunks` later on (see https://github.com/webpack/webpack/pull/9793#pullrequestreview-300371456).

Without it, the following error is thrown by the plugin:

```
TypeError: Cannot read property 'length' of undefined
    at (...)/node_modules/webpack-manifest-plugin/lib/plugin.js:128:39
```

Related: #186

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ node_js:
 
 env:
   - WEBPACK_VERSION=4
+  - WEBPACK_VERSION=next
 
 install:
   - npm install
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -77,6 +77,7 @@ ManifestPlugin.prototype.apply = function(compiler) {
         all: false,
         // Add asset Information
         assets: true,
+        ids: true, // needed by Webpack 5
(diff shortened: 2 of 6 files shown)


Lyrkan:
@mastilver Should I add `webpack@next` to the peer dependencies and tests? I noticed that you also changed them in #197, so maybe I should target that PR's branch instead?

mastilver:
@Lyrkan Yes, I believe your PR should target `next`  
Unfortunately I'm having some issue with the tests on that branch and until this is resolved I'm blocked

Lyrkan:
@mastilver Changed the target branch to `next` and tried to fix the tests after merging `master` to get the Webpack 5 fix that was commited on it (hence the additional commits and comment on this PR).

I also added back `webpack@next` to the CI matrix.

Hope I didn't forget anything :)

mastilver:
That's great, thank you

PR #205, opened 2019-11-06
Quick fix for #186

kirill-konshin:


mastilver:
please add a test also should probably target `next`

kirill-konshin:
Changed base to next. Can you suggest any existing test that I can take as example?

mastilver:
I think https://github.com/danethurber/webpack-manifest-plugin/blob/5715c07748a80dae583074304e306534eac1896a/spec/plugin.spec.js#L68 is a good place to start

Could you rebase on `next`, tests picking the right configuration

kirill-konshin:
The problem is that I have no idea how to force plugin in the state when `asset.chunks` is not defined.

mastilver:
Thank you, I just wanted to confirm it was an actual issue and it just happens that master started failing on this

PR #211, opened 2019-12-24
fix: recursively extract all assets from all child compilations 

### LINKED ISSUE #210: assets generated from the child compiler whose key has hash in the manifest.json
Currently, the `webpack-manifest-plugin` does not support extracting the asset chunk name from the child compiler.

Here is part of my `webpack.config.js`

```js
{
  ...
  "plugins": [
      new HtmlWebpackPlugin({
        "filename": "index.[chunkhash].html",
        "template": path.resolve(__dirname, "index.ejs")
      }),
      new ManifestPlugin()
  ],
  ...
}
```

And here is the actual output
```
{
  "index.js": "index.development.a26310385714a01e48eb.js",
  "vendors.js": "vendors.development.b4273a31cc345f8df775.js",
  "index.c5a9bff71fdfed9b6046.html": "index.c5a9bff71fdfed9b6046.html"
}
```

The expected output 
```
{
  "index.js": "index.development.a26310385714a01e48eb.js",
  "vendors.js": "vendors.development.b4273a31cc345f8df775.js",
  "index.html": "index.c5a9bff71fdfed9b6046.html"
}
```

In this output, how to we know the filename after the build without a stable key name.

And you will say that this is the issue on `html-webpack-plugin`. You are partially correct. Basically there are two issues right here. One is on `html-webpack-plugin` because it does not name the chunk. The other one is on `webpack-manifest-plugin`. Even if `html-webpack-plugin` name the chunk correctly, `webpack-manifest-plugin` won't be able to extract the chunk name as the key in the manifest.json

Because the key, `index.c5a9bff71fdfed9b6046.html`, was extracted from this logic https://github.com/danethurber/webpack-manifest-plugin/blob/b55ac5d7748dde89fd0225e13263c7b42b1dca87/lib/plugin.js#L115-L141. Child compiler can have the entry name. I think `webpack-manifest-plugin` should respect that if the chunks in child compiler have name.

The correct logic should recursively extract all the chunks from parent compilation to all children compilation

The following logic should be called recursively
https://github.com/danethurber/webpack-manifest-plugin/blob/b55ac5d7748dde89fd0225e13263c7b42b1dca87/lib/plugin.js#L87-L111

Please do not judge if the index.[contenthash].html makes sense. `worker-loader` is also a good counter example.

wood1986:
fix #210

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -84,59 +84,64 @@ ManifestPlugin.prototype.apply = function(compiler) {
         cachedAssets: true,
     });
 
-    var files = compilation.chunks.reduce(function(files, chunk) {
-      return chunk.files.reduce(function (files, path) {
-        var name = chunk.name ? chunk.name : null;
+    var evaluateAllCompilation = function (compilation, files) {
+      if (compilation.children.length) {
+        compilation.children.forEach(function (compilation) {
+          files = evaluateAllCompilation(compilation, files);
+        })
+      }
 
-        if (name) {
-          name = name + '.' + this.getFileType(path);
-        } else {
-          // For nameless chunks, just map the files directly.
-          name = path;
-        }
+      return compilation.chunks.reduce(function (files, chunk) {
+        return chunk.files.reduce(function (files, path) {
+          var name = chunk.name ? chunk.name : null;
+  
+          if (name) {
+            name = name + '.' + this.getFileType(path);
+          } else {
+            // For nameless chunks, just map the files directly.
+            name = path;
+          }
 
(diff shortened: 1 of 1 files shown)


shellscape:
@wood1986 Thanks for opening this PR a while back, and I'm sorry that it didn't get attention sooner. We've landed a major refactor in #222 that has resulted in some significant conflicts. Moving forward, we'll recommend the fix mentioned in https://github.com/shellscape/webpack-manifest-plugin/issues/210#issuecomment-641846388

Issue #137, opened 2018-04-06
Generate with compilation object


jscheid:
Hi, I'm looking after [webpack-subresource-integrity](https://github.com/waysact/webpack-subresource-integrity). We've had several requests for a feature that would let people export the per-asset integrity values as JSON, but I think a better solution might be to make it easy to use your plugin together with ours.

One simple solution would be if you could pass the `compilation` object as an additional argument to `generate`.  We'd then export a default generator function and advise people to use it like so:

```js
new ManifestPlugin({
  generate: SriPlugin.manifestGenerator(),
})
```

Our generator function would find the integrity values from the `compilation` object and generate something like `{ path: '...', integrity: '...' }` based on it.

What do you think?

mastilver:
I think that sounds great!! :)

Anything we should add from https://github.com/danethurber/webpack-manifest-plugin#hooks-options to help you implement that?

jscheid:
I guess if it had the original compilation asset that would be ideal. It would have to be the one from `compilation.assets`, not the one from `compilation.getStats().toJson().assets` because there isn't a good way yet for adding custom data to the stats JSON.

shellscape:
We've added support for the `beforeEmit` hook in f62ca8b273dcba5911e5f08814595c949120b6af. Please see the `Compiler Hooks` section in the README.

PR #216, opened 2020-03-27
Add reference to webpack.Compilation to generate

dirkgroenen:
For one of my projects I'm looking for a "manifest generator" which also adds the integrity of each asset. I've tried many of the available plugins, but none of them really seemed to be doing to job 100% correctly. 

Compared to the other available plugins I really like the quality and extendibility of this one. Especially the `generate` option is very useful and promising. 
In order to create the manifest I'm looking for I need some sort of reference to the Webpack compilation object, preferably from within the `generate` option. Unfortunately this isn't yet available, hence this PR.

Additionally I've also gone my way and added an example to help others solve the same problem. I've noticed there are quite some issues and topics around that try to cover this problem. 

This would also close https://github.com/danethurber/webpack-manifest-plugin/issues/137

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -23,7 +23,7 @@ function ManifestPlugin(opts) {
     map: null,
     generate: null,
     sort: null,
-    serialize: function(manifest) {
+    serialize: function (manifest) {
       return JSON.stringify(manifest, null, 2);
     },
   }, opts || {});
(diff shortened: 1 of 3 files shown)


rndm2:
Is this PR forgotten?

lib/plugin.js
```diff
@@ -23,7 +23,7 @@ function ManifestPlugin(opts) {
     map: null,
     generate: null,
     sort: null,
-    serialize: function(manifest) {
```
rndm2, line 26: I think you should revert changes in this file

shellscape:
Thanks for opening this PR a while back, and I'm sorry that it didn't get attention sooner. We've landed a major refactor in #222 that has resulted in some significant conflicts. We'll keep #137 open and will implement a fix along these lines, but will likely end up very different.

PR #226, opened 2020-11-05
Fix deprecated API calls and update tests

bungu:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [x] no

### Description

Some of the internal APIs were deprecated in webpack v5. There are two warnings:

```
(node:45864) [DEP_WEBPACK_COMPILATION_NORMAL_MODULE_LOADER_HOOK] DeprecationWarning: Compilation.hooks.normalModuleLoader was moved to NormalModule.getCompilationHooks(compilation).loader
    at getNormalModuleLoader (./node_modules/webpack/lib/Compilation.js:297:39)
    at Object.get normalModuleLoader [as normalModuleLoader] (./node_modules/webpack/lib/Compilation.js:603:12)
    at ./node_modules/webpack-manifest-plugin/lib/plugin.js:246:23
    ...
```

and

```
(node:45864) [DEP_WEBPACK_DEPRECATION_ARRAY_TO_SET] DeprecationWarning: Compilation.chunks was changed from Array to Set (using Array method 'reduce' is deprecated)
    at Set.set.<computed> [as reduce] (./node_modules/webpack/lib/util/deprecation.js:85:4)
    at ManifestPlugin.<anonymous> (./node_modules/webpack-manifest-plugin/lib/plugin.js:94:36)
   ...
```

After fixing warnings some of the tests started to fail. I think it was because of bug in webpack, but i'm not sure. It seems that webpack adds prefix `auto` to asset path, so manifest looks like this:

```json
{ "main.js": "automain.js" }
```

Probably this was caused by new placeholders `[base]` or `[path]` for file-based templates, because they have value `"auto"` by default.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -1,6 +1,7 @@
 var path = require('path');
 var fse = require('fs-extra');
 var _ = require('lodash');
+var NormalModule = require('webpack/lib/NormalModule');
 
 const emitCountMap = new Map();
 const compilerHookMap = new WeakMap();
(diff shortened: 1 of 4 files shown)


shellscape:
Thanks for putting this together. Because there are a number of critical bugs still open with webpack@5 I'm pausing one support for it until it stabilizes. I expect that to take another month or two. In the mean time, we'll leave this open in a hold state.

shellscape:
Just noticed that you committed into the `next` branch. Sorry for the confusion, but that branch is an orphan from the past maintainer and some past contributors. It's no longer being actively used and is there for archive and reference purposes. I'll rename it now to avoid future confusion. If you'd like to merge your changes, please rebase to master and reapply your changes there - warning: the codebase went through a full refactor and you'll have to reapply all changes on the new structure.

shellscape:
_(Also, please let me know if you would like me to apply your changes directly to master. Totally willing to do that as well)_

PR #246, opened 2021-01-17
Use same hook for both webpack 4 and 5

### LINKED ISSUE #245: Files generated by webpack plugins are not being added to the manifest file.
- Manifest Plugin Version: 3.0.0
- Webpack Version: 5.15
- Operating System (or Browser): Ubuntu 20
- Node Version: 14.15
- Link to reproduction : https://github.com/Levdbas/manifest-error

### Expected Behavior
when running npm run dev index.asset.json, generated by @wordpress/dependency-extraction-webpack-plugin, is also added to the manifest file.

### Actual Behavior
index.asset.json is missing from manifest file.

### Additional Information

I've the same issue with other webpack plugins that emit files as well, but I have limited this repo to only one example.
Nice detail: when setting webpack to 4.46 and webpack-manifest-plugin to 2.2.0 and retry to run npm run dev index.asset.json does get added to the manifest file.

So it seems like the files like this outside the entry files are not being added to the manifest. Any idea?

Levdbas:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no (using same tests as provided)

Breaking Changes?

- [x] no

List any relevant issue numbers: #245 

### Description

emitHook is now running on compiler.hooks.emit.tap(hookOptions, emit); for webpack 4/5. This fixes #245 
As said in the issue, I have little experience with webpack plugins, but the tests seems to work for both webpack 4 and 5.


Levdbas:
The CI tests are not running the webpack v5 testsuite but when running locally all 49 tests pass.

lib/index.js
```diff
@@ -57,13 +56,7 @@ class WebpackManifestPlugin {
...
     });
 
-    if (webpack.version.startsWith('4')) {
-      compiler.hooks.emit.tap(hookOptions, emit);
-    } else {
-      compiler.hooks.thisCompilation.tap(hookOptions, (compilation) => {
-        compilation.hooks.processAssets.tap(hookOptions, () => emit(compilation));
-      });
-    }
+    compiler.hooks.emit.tap(hookOptions, emit);
```
shellscape, line 59: so we put this different hook in for v5 based on direction from the webpack team itself. I'm actually surprised that the tests are passing with this change. I'll have to investigate recent changes in v5.x to see if something negated the need for the different hooks.

Levdbas: I see, I think you are referring to https://github.com/webpack/webpack/issues/11425 and the `DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated.
BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.` deprecation warning. It seems there are still a lot of plugins that are still using the emit hook and thus are called after `webpack-manifest-plugin` is done. So I think your code is the way to go but other plugins aren't ready yet.

Levdbas: Meanwhile I brought this incompatibility issue to the attention to the attention of the authors of the plugins from that I know of cause an issue with your updated code. See: https://github.com/WordPress/gutenberg/issues/27984#issuecomment-762100067 and https://github.com/roots/palette-webpack-plugin/issues/13.

So you are doing it right for sure but since you have done that before a lot of plugins are updated this now causes issues probably a lot of not yet updated plugins still using that old `emit` hook. 

Maybe it is possible to make something like a check function at the end of the compilation that gives a warning if other plugins are adding assets too late and thus are not added to the manifest?

shellscape: I gave this one some serious thought. I see where the issues for your builds are valid and aren't going away anytime soon - yet another symptom of adopting new webpack majors within the first year they're available. There's never been a smooth webpack major release, they're always a mess.

I think the right thing to do here is to introduce a new option: `useLegacyEmit`, which will use the webpack v4 emit pattern even for webpack v5. It should be `false` by default, and the documentation for that option should state explain the scenario that you're facing and why the option is needed, and should also warn users that the `DeprecationWarning` will be visible. That should maintain compatibility with the folks relying on this plugin to emit in the new patterns, as well as allow folks to work around that.

Levdbas: Thank you for your reply. I do think this is a very good middle ground for now! Maybe it's a good idea to write a standard text that users can utilize to make an issue at other plugin repositories to point authors at said issue and a way to fix it in their webpack plugins. I am willing to write a first version if you'd like to.

Levdbas:
PR now contains the new  `useLegacyEmit` option.  I also added a small readme text. It might be smart to add a test but I've tested it locally and it just works. Since my knowledge about ava is non-existent I hope you can take a look at that.

Circleci fails on the security audit btw.

shellscape:
@Levdbas please merge from upstream/master for the audit fix. We're going to need a test specifically for using webpack v5 with the plugin and `useLegacyEmit: true` before I can merge this. Using the existing tests aren't sufficient to cover that scenario.

Levdbas:
Upstream is merged. Any idea how to make this test work? Include a plugin from which we know only emits on the legacy hook and lock the version? I have no experience with AVA yet so I will see what I can do.

shellscape:
Ava is very similar to Jest, the learning curve is small. That sounds like a good test plan to me.

Levdbas:
Well that was fairly straightforward. Let me know if everything is in order this way.

Levdbas:
Hi @shellscape , did you had the chance to take a look at the test?

Levdbas:
Hi @shellscape , any word if and when you can merge this and release this in a new version?

shellscape:
Yeah this PR is going to be part of the last release of support for webpack v4 and Node v10. Stand by.

privatenumber:
I'd like to chime in and say that I'm troubled by the hook discrepancy as well.

I wrote a [localization plugin](https://github.com/privatenumber/webpack-localize-assets-plugin), which duplicates assets and localizes them (essentially just string replacement) _after the optimization/minification stage_ (because we don't want minification to repeat for the duplicated assets).

For WP4, the localization runs in the [`afterOptimizeChunkAssets` sync hook](https://github.com/privatenumber/webpack-localize-assets-plugin/blob/2ec43fc5e2b8953115aaa87ffb0e0ecf824e98fb/src/index.ts#L322), which works fine with `webpack-manifest-plugin` because it runs on the `compiler.emit` hook which comes afterwards.

For WP5, the localization runs in the [`afterProcessAssets` hook](https://github.com/privatenumber/webpack-localize-assets-plugin/blob/2ec43fc5e2b8953115aaa87ffb0e0ecf824e98fb/src/index.ts#L317), which doesn't work with `webpack-manifest-plugin` because `webpack-manifest-plugin` runs at the end of the `processAssets` hook.

The reason why I can't hook into `processAssets` in WP5 is because minifiers can enable the [`additionalAssets` option](https://webpack.js.org/api/compilation-hooks/#additional-assets) to re-run on new assets emitted by other plugins in `processAssets`. (eg. [Terser](https://github.com/webpack-contrib/terser-webpack-plugin/blob/bfbb68a7455651990a384b29869924a7bf57fade/src/index.js#L656), [esbuild-loader](https://github.com/privatenumber/esbuild-loader/blob/9f6065ac01757dd49baab1931ac16fc5aca89758/src/minify-plugin.ts#L89)).

That said, I might recommend not using the `processAssets` hook for this. I think `compilation.afterProcessAssets` or `compiler.emits` (via unsealing and re-sealing) makes more sense for this plugin. Alternatively, you can also tap into the [`record` hook](https://github.com/webpack/webpack/blob/226a77c9d46b33da5b78b1c76a10384c78132074/lib/Compilation.js#L2511) which runs after assets are processed and right before the compilation is sealed.

----

> so we put this different hook in for v5 based on direction from the webpack team itself.

@shellscape Curious what reasoning they gave you?

shellscape:
@privatenumber I honestly can't remember now what the feedback was, it's just been too long. One of those "I remember why, but not what" situations. That said, I'm all for using `afterProcessAssets`. Sounds completely reasonable to me. If you'd like to open a PR to make that change, I'd be happy to accept it (pending passing CI, etc of course). I'm going to merge this one for the time being, which should help some folks out.

PR #249, opened 2021-02-07
Fixing extra and invalid paths in Webpack 5

### LINKED ISSUE #237: Manifest populates main entry point with unrelated files with webpack 5
- Manifest Plugin Version: 3.0.0
- Webpack Version: 5.6.0
- Operating System (or Browser): Chrome 87 / Win 10 64-bit
- Node Version: 12.16.1
- Link to reproduction _(⚠️ read below)_: https://repl.it/@KrofDrakula/manifest-plugin-repro

### Expected Behavior

When configuring external files via `file-loader` in the `rules` section, I would not expect the files to populate the manifest file with the entry name and the imported file. In the above example I would expect:

```json
{
  "main.js": "main-c53637abe0.js",
  "demo.svg": "demo-d96fd782eb31f2e1c11b1a3f0ab2bb18.svg"
}
```

### Actual Behavior

Importing an `.svg` file using `file-loader` with content hashes causes the manifest to override the entry name and appending the loaded file extension, in my case:

```json
{
  "main.js": "main-c53637abe0.js",
  "main.svg": "demo-d96fd782eb31f2e1c11b1a3f0ab2bb18.svg",
  "demo.svg": "demo-d96fd782eb31f2e1c11b1a3f0ab2bb18.svg"
}
```

### Additional Information

My guess is that this has to do with the plugin matching the content hash and assuming the loaded assets are also entry points.

### LINKED ISSUE #238: Manifest keys contain source path
- Manifest Plugin Version: 3.0.0
- Webpack Version: 5.11.0
- Reproduction: https://github.com/bigskillet/webpack-manifest-plugin

### Expected Behavior
```
  "/assets/main.css": "/assets/main-fed4ead9.css",
  "/assets/main.js": "/assets/main-31d6cfe0.js",
  "/assets/test.png": "/assets/test-7f730f47.png",
  "/assets/font.woff2": "/assets/font.woff2"
```

### Actual Behavior
```
  "/assets/main.css": "/assets/main-fed4ead9.css",
  "/assets/main.js": "/assets/main-31d6cfe0.js",
  "/assets/assets/test.png": "/assets/test-7f730f47.png",
  "/assets/index.html": "/index.html",
  "/assets/src/assets/static/font.woff2": "/assets/font.woff2"
```

### Additional Information
The manifest plugin is adding the source path the entry key. For example, `/src/assets/static/` is being added to the copied static font file. It's not adding the source path to the main.css and main.js keys though. I had to add a `basePath` to get them matching, which caused the test.png key to double the `/assets/` path and also added it to the index.html key. If I remove the `basePath`, the main entries don't match up:

- `"main.css": "/assets/main-fed4ead9.css"`
- `"main.js": "/assets/main-31d6cfe0.js"`

Sorry, I'm a bit of n00b, so I'm not sure if this is a bug, feature, or if I'm just using it wrong.

Thanks!

weaverryan:
This PR contains:

- [X] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [X] yes
- [] no

Breaking Changes?

- [X] no

List any relevant issue numbers: Fixes #238 and fixes #237 (also fixes https://github.com/symfony/webpack-encore/issues/907)

### Description

Hi!

I maintain @symfony/webpack-encore, and we recently upgraded to Webpack 5 and have hit issues with the `manifest.json` file. So, I dug into it :). 

This is complex, and I hope someone may have some ideas on how to make this nicer. Here is the situation:

A) As far as I (and the test suite) can tell, `chunk.auxiliaryFiles` is *only* needed to get sourcemap files. All other files are available in other ways. Since *all* `chunk.auxiliaryFiles` were used before, it caused #237 because when we loop over these files, we re-use the same `chunk.name` over and over again (and then just add the correct extension like `.svg`). So, we need to use *some* of these, but not all of them. To do that, we collect all the "auxiliary files", then check which have *not* been added to the manifest at the end. Those that are missing (sourcemaps) are added.

B) The second issue - and the one that fixes #238 - is more complex. Basically, if you're using the new "asset modules" system in Webpack 5, then the `emitFile()` method - which we rely on in the `normalModuleLoaderHook` is never called. In fact, as far as I can tell, this method was *only* ever called by the `file-loader` - https://github.com/webpack-contrib/file-loader/blob/467e6b782bb8bb9fcb54dcb981b5df1a2aeae818/src/index.js#L81 (`this` is the `loaderContext`) - it's never called by Webpack directly.

When you use "asset modules", there is no `file-loader` and so this is never called. According to the Webpack people - https://github.com/webpack/webpack/issues/12637#issuecomment-776185604 - we should rely on `asset.info.sourceFilename` as the "source of truth". We were already using that, but sort of "incorrectly" - the `asset.info.sourceFilename` will be equal to something like `'../../assets/logo.png'` or `'node_modules/something/fonts/lato-normal.abc123.woff2'`. So, what we need is the `basename` of this prefixed by the `dirname` of `asset.name` (which will be something like `fonts/lato-normal.abc123.woff2`).

## To test in your project

If you're having this issue, I would love if this fix could be validated:

```
yarn remove webpack-manifest-plugin
yarn add git://github.com/weaverryan/webpack-manifest-plugin.git#fix-invalid-paths-webpack5
```

Thanks!

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -26,14 +26,12 @@ const getFileType = (fileName, { transformExtensions }) => {
   return transformExtensions.test(extension) ? `${split.pop()}.${extension}` : extension;
 };
 
-const reduceAssets = (files, asset, moduleAssets, assetTypeModuleAssets) => {
+const reduceAssets = (files, asset, moduleAssets) => {
   let name;
   if (moduleAssets[asset.name]) {
     name = moduleAssets[asset.name];
-  } else if (assetTypeModuleAssets[asset.info.sourceFilename]) {
-    name = join(dirname(asset.name), assetTypeModuleAssets[asset.info.sourceFilename]);
-  } else {
-    name = asset.info.sourceFilename;
+  } else if (asset.info.sourceFilename) {
+    name = join(dirname(asset.name), basename(asset.info.sourceFilename));
   }
 
   if (name) {
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,3 +1,5 @@
+const { dirname, join, basename } = require('path');
+
 const generateManifest = (compilation, files, { generate, seed = {} }) => {
   let result;
   if (generate) {
@@ -24,8 +26,16 @@ const getFileType = (fileName, { transformExtensions }) => {
   return transformExtensions.test(extension) ? `${split.pop()}.${extension}` : extension;
 };
 
-const reduceAssets = (files, asset, moduleAssets) => {
-  const name = moduleAssets[asset.name] ? moduleAssets[asset.name] : asset.info.sourceFilename;
+const reduceAssets = (files, asset, moduleAssets, assetTypeModuleAssets) => {
+  let name;
+  if (moduleAssets[asset.name]) {
+    name = moduleAssets[asset.name];
+  } else if (assetTypeModuleAssets[asset.info.sourceFilename]) {
+    name = join(dirname(asset.name), assetTypeModuleAssets[asset.info.sourceFilename]);
+  } else {
+    name = asset.info.sourceFilename;
+  }
+
   if (name) {
     return files.concat({
       path: asset.name,
@@ -52,29 +62,43 @@ const reduceAssets = (files, asset, moduleAssets) => {
   });
 };
 
-const reduceChunk = (files, chunk, options) =>
-  Array.of(...Array.from(chunk.files), ...Array.from(chunk.auxiliaryFiles || [])).reduce(
-    (prev, path) => {
-      let name = chunk.name ? chunk.name : null;
-      // chunk name, or for nameless chunks, just map the files directly.
-      name = name
-        ? options.useEntryKeys && !path.endsWith('.map')
-          ? name
-          : `${name}.${getFileType(path, options)}`
-        : path;
+const reduceChunk = (files, chunk, options, auxiliaryFiles) => {
+  // auxiliary files contain things like images, fonts AND, most
+  // importantly, other files like .map sourcemap files
+  // we modify the auxiliaryFiles so that we can add any of these
+  // to the manifest that was not added by another method
+  // (sourcemaps files are not added via any other method)
+  Array.from(chunk.auxiliaryFiles || []).forEach((auxiliaryFile) => {
+    auxiliaryFiles[auxiliaryFile] = {
+      path: auxiliaryFile,
+      name: basename(auxiliaryFile),
+      isInitial: false,
+      isChunk: false,
+      isAsset: true,
+      isModuleAsset: true
+    };
+  });
 
-      return prev.concat({
-        path,
-        chunk,
-        name,
-        isInitial: chunk.isOnlyInitial(),
-        isChunk: true,
-        isAsset: false,
-        isModuleAsset: false
-      });
-    },
-    files
-  );
+  return Array.from(chunk.files).reduce((prev, path) => {
+    let name = chunk.name ? chunk.name : null;
+    // chunk name, or for nameless chunks, just map the files directly.
+    name = name
+      ? options.useEntryKeys && !path.endsWith('.map')
+        ? name
+        : `${name}.${getFileType(path, options)}`
+      : path;
+
+    return prev.concat({
+      path,
+      chunk,
+      name,
+      isInitial: chunk.isOnlyInitial(),
+      isChunk: true,
+      isAsset: false,
+      isModuleAsset: false
+    });
+  }, files);
+};
 
 const standardizeFilePaths = (file) => {
   const result = Object.assign({}, file);
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -1,5 +1,5 @@
 const { mkdirSync, writeFileSync } = require('fs');
-const { basename, dirname, join, relative } = require('path');
+const { basename, dirname, join } = require('path');
 
 const { SyncWaterfallHook } = require('tapable');
 const webpack = require('webpack');
@@ -33,15 +33,7 @@ const beforeRunHook = ({ emitCountMap, manifestFileName }, compiler, callback) =
 };
 
 const emitHook = function emit(
-  {
-    compiler,
-    emitCountMap,
-    manifestAssetId,
-    manifestFileName,
-    moduleAssets,
-    assetTypeModuleAssets,
-    options
-  },
+  { compiler, emitCountMap, manifestAssetId, manifestFileName, moduleAssets, options },
   compilation
 ) {
   const emitCount = emitCountMap.get(manifestFileName) - 1;
@@ -66,10 +58,7 @@ const emitHook = function emit(
   );
 
   // module assets don't show up in assetsByChunkName, we're getting them this way
-  files = stats.assets.reduce(
-    (prev, asset) => reduceAssets(prev, asset, moduleAssets, assetTypeModuleAssets),
-    files
-  );
+  files = stats.assets.reduce((prev, asset) => reduceAssets(prev, asset, moduleAssets), files);
 
   // don't add hot updates and don't add manifests from other instances
   files = files.filter(
@@ -136,23 +125,8 @@ const emitHook = function emit(
   getCompilerHooks(compiler).afterEmit.call(manifest);
 };
 
-const normalModuleLoaderHook = ({ moduleAssets, assetTypeModuleAssets }, loaderContext, module) => {
+const normalModuleLoaderHook = ({ moduleAssets }, loaderContext, module) => {
   const { emitFile } = loaderContext;
-
-  // the "emitFile" callback is never called on asset modules
-  // so, we create a different map that can be used later in the "emit" hook
-  if (['asset', 'asset/inline', 'asset/resource', 'asset/source'].includes(module.type)) {
-    // This takes the userRequest (which is an absolute path) and turns it into
-    // a relative path to the root context. This is done so that the string
-    // will match asset.info.sourceFilename in the emit hook.
-    let sourceFilename = relative(loaderContext.rootContext, module.userRequest);
-    // at this point, Windows paths use \ in their paths
-    // but in the emit hook, asset.info.sourceFilename fill have UNIX slashes
-    sourceFilename = sourceFilename.replace(/\\/g, '/');
-    Object.assign(assetTypeModuleAssets, {
-      [sourceFilename]: basename(module.userRequest)
-    });
-  }
 
   // eslint-disable-next-line no-param-reassign
   loaderContext.emitFile = (file, content, sourceMap) => {
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -1,5 +1,5 @@
 const { mkdirSync, writeFileSync } = require('fs');
-const { basename, dirname, join } = require('path');
+const { basename, dirname, join, relative } = require('path');
 
 const { SyncWaterfallHook } = require('tapable');
 const webpack = require('webpack');
@@ -33,7 +33,15 @@ const beforeRunHook = ({ emitCountMap, manifestFileName }, compiler, callback) =
 };
 
 const emitHook = function emit(
-  { compiler, emitCountMap, manifestAssetId, manifestFileName, moduleAssets, options },
+  {
+    compiler,
+    emitCountMap,
+    manifestAssetId,
+    manifestFileName,
+    moduleAssets,
+    assetTypeModuleAssets,
+    options
+  },
   compilation
 ) {
   const emitCount = emitCountMap.get(manifestFileName) - 1;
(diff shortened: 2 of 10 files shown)


shellscape:
I would absolutely love to accept this PR, but without tests to guard against regressions I can't. (I do see that you state that tests are `TODO`). 

> I believe chunk.auxiliaryFiles should have never been included when getting the file list

The use of `auxillaryFiles` was introduced here https://github.com/shellscape/webpack-manifest-plugin/commit/2702efb04a0c54da7ce17a082c5022e6ad9e71e6 to help with v5 compat _at that time_. The moving goalposts of v5 in the early minor versions made it hard to keep up with, as per usual with new webpack majors. It's cool if that's no longer needed, but we'll need to guard against regressions there.

lib/helpers.js
```diff
@@ -52,29 +62,43 @@ const reduceAssets = (files, asset, moduleAssets) => {
...
+      path: auxiliaryFile,
+      name: basename(auxiliaryFile),
+      isInitial: false,
+      isChunk: false,
+      isAsset: true,
+      isModuleAsset: true
+    };
+  });
+
+  return Array.from(chunk.files).reduce((prev, path) => {
```
weaverryan, line 83: If you view this section without whitespace - https://github.com/shellscape/webpack-manifest-plugin/pull/249/files?w=1 - you'll see that the existing logic remains the same. The `chunk.auxiliaryFiles` were removed from the "normal" logic and moved into the above loop to track those custom.

lib/helpers.js
```diff
@@ -52,29 +62,43 @@ const reduceAssets = (files, asset, moduleAssets) => {
...
+  // to the manifest that was not added by another method
+  // (sourcemaps files are not added via any other method)
+  Array.from(chunk.auxiliaryFiles || []).forEach((auxiliaryFile) => {
+    auxiliaryFiles[auxiliaryFile] = {
+      path: auxiliaryFile,
+      name: basename(auxiliaryFile),
+      isInitial: false,
+      isChunk: false,
+      isAsset: true,
+      isModuleAsset: true
```
weaverryan, line 79: I believe these are all correct: these are not chunks, they are assets. And I believe they are "module assets"  because they are auxiliary assets to a chunk, but I'm not positive if that is the intention. These will most notably include sourcemap files.

lib/hooks.js
```diff
@@ -33,7 +33,15 @@ const beforeRunHook = ({ emitCountMap, manifestFileName }, compiler, callback) =
...
-  { compiler, emitCountMap, manifestAssetId, manifestFileName, moduleAssets, options },
+  {
+    compiler,
+    emitCountMap,
+    manifestAssetId,
+    manifestFileName,
+    moduleAssets,
+    assetTypeModuleAssets,
+    options
+  },
```
weaverryan, line 44: A few of these cs changes are being made automatically by the linter - I tried to keep the diff as clean as possible :)

test/integration/location.js
```diff
@@ -22,12 +22,12 @@ test('output to the correct location', async (t) => {
       filename: '[name].js',
       path: outputPath
     },
-    plugins: [new WebpackManifestPlugin({ fileName: 'webpack.manifest.js' })]
+    plugins: [new WebpackManifestPlugin({ fileName: 'webpack.manifest.json' })]
```
weaverryan, line 25: If you end the file in `.js`, the TerserPlugin tries to parse these files and explodes. I'm not sure why this is different between v4 and v5 (it's only a problem on v5), but it makes sense: the manifest contents are not actually a valid JavaScript file (`{ "foo": "bar" }` is not valid JavaScript - you'd need to have a `const = ` in front or something equal).

And so, I think this is valid to change. Users are able to control the filename (the point of the test), but in Webpack 5, they cannot name it `.js`.

test/unit/index.js
```diff
@@ -75,15 +74,15 @@ test('works with source maps', async (t) => {
       one: '../fixtures/file.js'
     },
     output: {
-      filename: '[name].js',
+      filename: 'build/[name].js',
```
weaverryan, line 78: Just making the test more interesting.

test/unit/index.js
```diff
@@ -158,11 +157,6 @@ test('outputs a manifest of no-js file', async (t) => {
     'file.txt': 'file.txt'
   };
 
-  // Note: I believe this to be another bug in webpack v5 and cannot find a good workaround atm
-  if (webpack.version.startsWith('5')) {
-    expected['main.txt'] = 'file.txt';
-  }
```
weaverryan, line 164: It turns out, these are now fixed! They are #237

lib/helpers.js
```diff
@@ -24,8 +26,17 @@ const getFileType = (fileName, { transformExtensions }) => {
   return transformExtensions.test(extension) ? `${split.pop()}.${extension}` : extension;
 };
 
-const reduceAssets = (files, asset, moduleAssets) => {
-  const name = moduleAssets[asset.name] ? moduleAssets[asset.name] : asset.info.sourceFilename;
+const reduceAssets = (files, asset, moduleAssets, assetTypeModuleAssets) => {
+  console.log(assetTypeModuleAssets, asset.info.sourceFilename);
```
shellscape, line 30: This little guy snuck into the party

weaverryan: Oh yea :) I'm actively debugging a last Windows issue that I can only replicate on this CI. So i'm abusing the Windows CI (I don't have access to any other Windows machine). 

I'll let you know when things are ready to look at - I thought I had it so I must be missing a tiny last detail :)

test/integration/location.js
shellscape: Is that something we should document in the README?

weaverryan:
I'm stuck on one last detail for now: for some reason, in Windows only, `asset.info.sourceFilename` in `reduceAssets()` - https://github.com/shellscape/webpack-manifest-plugin/blob/master/lib/helpers.js#L28 - is always missing.

What's stranger is that on our CI - we use AppVeyor - it is *not* missing, and everything works fine. I have no idea why that would be the case :/.

shellscape:
Are there any other environment differences on AppVeyor versus our Actions? Node versions, other installed windows weird things?

weaverryan:
Hmm, i've actually just added the same Windows matrix to our PR, and it's working fine - https://github.com/symfony/webpack-encore/pull/921/checks#step:6:11 - that test relies on the same thing that is mysteriously missing here. We're executing Webpack in exactly the same way... 🤔

weaverryan:
I've asked for help here: https://github.com/webpack/webpack/issues/12637

I absolutely had this issue cornered from all angles... until I hit the Windows bug in this CI only (which I can't repeat anywhere else, not even on my own CI). In general, what this plugin tries to do is simple: get a list of all files that were emitted and make a map from the `"sourceFilename": "outputFilename"` in JSON. On a high level, that should be simple to do. But this library has to do all kinds of tricks to do it: getting the `.js` and `.css` entry files from `compilation.chunks`, hooking into `file-loader` via the `emitFile()` callback, auxiliary files and (now) hooking into `NormalModule.loader` for asset modules.

Anyways, I'm hoping there is a simpler way. When `asset.info.sourceFilename` vanished from images on the Windows CI here only, that sort of broke all my plans and I have not found a way to make the map from "output filename" to "source filename" any other way.

If you have any ideas, I would love to hear them :).

Thanks!

test/integration/location.js
weaverryan: Hmm. I would say, let's wait to see if we have any questions about it. The only reason I'm hesitating is that the JSON was *never* a valid JavaScript file... so I'm trying to think of what the use-case ever would have been (and so, is anyone out there actually doing that?).

lib/helpers.js
```diff
@@ -25,7 +27,13 @@ const getFileType = (fileName, { transformExtensions }) => {
 };
 
 const reduceAssets = (files, asset, moduleAssets) => {
-  const name = moduleAssets[asset.name] ? moduleAssets[asset.name] : asset.info.sourceFilename;
+  let name;
+  if (moduleAssets[asset.name]) {
+    name = moduleAssets[asset.name];
+  } else if (asset.info.sourceFilename) {
+    name = join(dirname(asset.name), basename(asset.info.sourceFilename));
```
weaverryan, line 34: This is basically the same thing that's already being done at the bottom of `hooks.js` for the `file-loader`/ NormalModule situation - https://github.com/shellscape/webpack-manifest-plugin/blob/9f408f609d9b1af255491036b6fc127777ee6f9a/lib/hooks.js#L122

set-webpack-version.js
```diff
@@ -0,0 +1,25 @@
+/*
+ * This file sets Webpack version - both devDependencies and peerDependencies.
+ *
+ * This is, for some reason, needed with Windows and maybe npm 6. Running
+ * "npm install webpack -D" AND manually updating the peerDependencies
+ * (because npm 6 does not do that but npm 7 does) is not enough. For some
+ * reason why must, by hand, update the package.json file first and then
+ * run a normal "npm install".
+ */
```
weaverryan, line 9: This is insanity, but THIS turned out to be the reason that `asset.info.sourceFilename` was missing on Windows (and why I could repeat it). I know from the Webpack folks - https://github.com/webpack/webpack/issues/12637#issuecomment-776185604 - that relying on `asset.info.sourceFilename` IS the correct way to do things.

Basically, for reasons that aren't clear, on Windows only, if you installed Webpack 5 via `npm install webpack -D` it would *not* work. Somewhere, internally, some dependencies must have gotten "crossed". But if you manually change the `peerDependencies` and `devDependencies` for Webpack and then run `npm install`, things work fine. This... almost killed me. But it turns out that this was an edge-case npm + Windows bug of some sort... or possibly some edge case since the `package.json` file is mixed with libraries that have Webpack 4 vs Webpack 5 as their peer dependency.

test/unit/index.js
```diff
@@ -188,3 +183,37 @@ test('make manifest available to other webpack plugins', async (t) => {
...
+    const { manifest } = await compile(config, t);
+    const expected = {
+      'main.js': 'main.js',
+      'images/manifest.svg': `images/manifest.14ca.svg`
+    };
+
+    t.truthy(manifest);
+    t.deepEqual(Object.keys(expected), ['main.js', 'images/manifest.svg']);
+    t.deepEqual(manifest['main.js'], 'main.js');
+    t.regex(manifest['images/manifest.svg'], /images\/manifest\.[a-z|\d]{4}\.svg/);
```
weaverryan, line 215: Doing this so I could do a fuzzy check on the hash. The `[fullhash]` thing does not work for `assetModuleFilename`, so I needed to use a *real* hash.

weaverryan:
This is now ready!!!!!!

I have updated the PR description. In short, a crazy npm+Windows bug caused an "invalid" problem, which the test suite now works around. I was also able to simplify the PR a bit after getting guidance from the Webpack folds.

Please let me know if you need anything else - I'm hoping (now that this is finally complete... and it almost killed me 😉 ) that we could release this soon (and then I can avoid releasing my "copy" of the plugin inside symfony/webpack-encore).

Thanks!

set-webpack-version.js
shellscape: This is fascinating. What we need to know before we merge is why this was an issue only _now_. We've been testing webpack v4 and webpack v5 separately with the windows workflows without issue up to this point. So what's the missing bit of nuance as to why it's an issue now?

shellscape: @weaverryan ping on this comment

aboks:
@weaverryan as we were having some issues with Webpack 5, webpack-manifest-plugin and copy-webpack-plugin (seemingly related to #237 and #238), I tried out your branch. It seems to fix some of the issues, but still handles one case incorrectly. 

We have something in our Webpack config that looks like
```
new CopyWebpackPlugin({
    "patterns": [
        { "from": "./path/to/common/favicon.ico", "to": "favicon.ico" },
        { "from": "./path/to/a/favicon.ico", "to": "favicon_a.ico" },
        { "from": "./path/to/b/favicon.ico", "to": "favicon_b.ico" },
    ]
})
```

Under Webpack 4 (with webpack-manifest-plugin 2.2.0) this resulted in the following manifest entries. I assume these to be correct.
```
{
    "favicon.ico": "favicon.ico",
    "favicon_a.ico": "favicon_a.ico",
    "favicon_b.ico": "favicon_b.ico"
}
```

Under Webpack 5 (with webpack-manifest-plugin 3.0.0, without your fix) this resulted in the following entries:
```
{
    "path/to/common/favicon.ico": "favicon.ico",
    "path/to/a/favicon.ico": "favicon_a.ico",
    "path/to/b/favicon.ico": "favicon_b.ico"
}
```

Now with your fix we only get a single manifest entry for the favicons:
```
{
    "favicon.ico": "favicon_b.ico"
}
```
It seems like the logic to combine the dirname from one path with the basename of another from https://github.com/weaverryan/webpack-encore/commit/d83e6dbe43fbaa911e2364b24c14f77bcc331dab#diff-8a4026eedbd67006d86faab015b15c08372284c5c957cfd5a49e0faf8dd85396R34 does not work nicely when files copied using copy-webpack-plugin get a different name than they had originally, especially if the original names collide.

Would there be a way to change your fix to handle this situation correctly? I looked at a possible solution, but I have to admit I'm too unfamiliar with Webpack internals and asset modules to grasp what exactly is going on. Maybe it's necessary to distinguish the asset module case from the copied asset case using `asset.info.contenthash` or `asset.info.copied`?

shellscape:
@aboks I'm going to merge this one now so folks get the benefit of the changes. We can iterate to solve that last case later on.

shellscape:
thanks!

Issue #270, opened 2021-07-27
peerDependency on webpack broken for webpack 5 since v3.2.0


jorrit:
In commit 79987fcd1314286a33441d306e7b85b5445056b2, which is included in 3.2.0, the peerDependency argument of webpack got changed from `>=4.44.2` to `^4.44.2`. This means that from 3.2.0 on, Webpack 5 is not allowed as a peer dependency.

- Manifest Plugin Version: 3.2.0
- Webpack Version: 5.46.0
- Operating System (or Browser): Win 10
- Node Version: 14
- Link to reproduction _(⚠️ read below)_:

### Repro

With NPM 7:

1. cd to some empty directory
2. `npm install webpack@5`
3. `npm install webpack-manifest-plugin`

### Expected Behavior

Successful installation

### Actual Behavior

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: undefined@undefined
npm ERR! Found: webpack@5.46.0
npm ERR! node_modules/webpack
npm ERR!   webpack@"^5.46.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer webpack@"^4.44.2" from webpack-manifest-plugin@3.2.0
npm ERR! node_modules/webpack-manifest-plugin
npm ERR!   webpack-manifest-plugin@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

### Additional information

Running `npm install webpack-manifest-plugin@3.1.1` works fine.

ali-hosseini-fs:
When do you want to publish @jorrit s PR to npm?
I updated npm/node version on our project and because of this problem on your package I can't continue my work on it!
I will appreciate if you can publish it as soon as possible.

shellscape:
@ali-hosseini-fs when I'm damn well ready to. there isn't a maintainer alive who appreciates those kinds of replies. stop it.

shellscape:
Closed by #271 _yesterday_

yura3d:
Just updated to 4.0.0 and still see warning for incorrect webpack version (^4.44.2) as peer dependency. Opened package.json and saw:
[...]

PR #271, opened 2021-07-27
Allow webpack 5 as peer dependency

jorrit:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #270

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "LICENSE"
   ],
   "peerDependencies": {
-    "webpack": "^4.44.2"
+    "webpack": "^4.44.2 || ^5.46.0"
   },
   "dependencies": {
     "tapable": "^2.0.0",


package.json
```diff
@@ -21,7 +21,7 @@
     "lint:js": "eslint --fix --cache lib test",
     "lint:json": "prettier --write codecov.yml .circleci/config.yml .eslintrc",
     "lint:package": "prettier --write package.json --plugin=prettier-plugin-package",
-    "posttest": "node set-webpack-version.js \"^4.44.2\" && npm install",
+    "posttest": "node set-webpack-version.js \">=4.44.2\" && npm install",
```
shellscape, line 24: please revert this change

package.json
```diff
@@ -33,7 +33,7 @@
     "LICENSE"
   ],
   "peerDependencies": {
-    "webpack": "^4.44.2"
+    "webpack": ">=4.44.2"
```
shellscape, line 36:
```suggestion
    "webpack": "^4.44.2 || ^5.46.0"
```

Using `>=` would allow v6 if/when that happens, which is not correct

shellscape (changes requested):

package.json
jorrit: Should this become '^4.44.2 || ^5.46.0' just like the other setting? Because if I simply revert it, testing will alter package.json and perhaps cause committing unwanted changes.

package.json
jorrit: Sure thing. My initial solution was to revert to the previous setting, but `^4.44.2 || ^5.46.0` is better.

package.json
shellscape: There's a larger fix here that's needed related to the set-webpack-version script to prevent bad overwrites. I'll do that after your PR.

jorrit: Cool

shellscape (approved):

jorrit:
This PR was reverted by 830a1b20d0de5e661b0f51cef62c0afa0945700b.

shellscape:
It actually wasn't. It was reverted by the previous commit, and reinstated by the commit you referenced.

jorrit:
Sorry, I shouldn't have commented before the third coffee in the morning. My apologies.

PR #297, opened 2022-12-19
fix: only import webpack types

goloveychuk:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

So I've somehow messed with dependency tree (with monorepo+linking) and webpack loaded twise. 
While you have setup webpack as peerDep and everything should be ok, I guess better to get webpack module provided instead of importing it. Also easier ts types

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,10 +1,7 @@
 import { relative, resolve } from 'path';
 
 import { SyncHook } from 'tapable';
-import { Compiler, WebpackPluginInstance, Compilation } from 'webpack';
-// @ts-ignore
-import NormalModule from 'webpack/lib/NormalModule';
-
+import type { Compiler, WebpackPluginInstance, Compilation } from 'webpack';
 import { FileDescriptor } from './helpers';
 import { beforeRunHook, emitHook, getCompilerHooks, normalModuleLoaderHook } from './hooks';
 
@@ -71,6 +68,7 @@ class WebpackManifestPlugin implements WebpackPluginInstance {
   }
 
   apply(compiler: Compiler) {
+    const { NormalModule } = compiler.webpack;
     const moduleAssets = {};
     const manifestFileName = resolve(compiler.options.output?.path || './', this.options.fileName);
     const manifestAssetId = relative(compiler.options.output?.path || './', manifestFileName);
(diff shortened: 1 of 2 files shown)


goloveychuk:
reapprove please

shellscape:
Please rebase from upstream/master

src/index.ts
```diff
@@ -93,7 +91,7 @@ class WebpackManifestPlugin implements WebpackPluginInstance {
       const hook = !NormalModule.getCompilationHooks
         ? compilation.hooks.normalModuleLoader
         : NormalModule.getCompilationHooks(compilation).loader;
-      hook.tap(hookOptions, normalModuleLoader);
+      hook.tap(hookOptions, normalModuleLoader as any);
```
shellscape, line 94: The use of `as any` here will be a blocker for this PR

goloveychuk: so without casting it won't work

shellscape: look at the types before your changes (what's on master), and compare to what it is with your changes, adjust from there.

goloveychuk: before my change it has import
```ts
// @ts-ignore
import NormalModule from 'webpack/lib/NormalModule';
```
so NormalModule was any. 
You suggest to replicate previous types, so I'll need to 
```ts
    const NormalModule = compiler.webpack.NormalModule as any;
```

goloveychuk: I think upcasting from object to more specific type inside normalmodule hook is better, more types are preserved/checked

goloveychuk: > second try. 
By this I meant that I've made second fix, removed as any.

shellscape:
made more updates to upstream/master. webpack needed an upgrade to support node v18 in the tests.

goloveychuk:
rebased

goloveychuk:
I think there is no better fix, wdyt?

wlblanchette:
Wanted to drop a comment here that this fix solves an instance of `TypeError: The 'compilation' argument must be an instance of Compilation` that results from a monorepo environment where the webpack configuration is defined in one package, with one instance of `webpack` and then imported into a second package where, for whatever reason, the second package its own instance of webpack.

In our case, it was a symptom of hoisting falling over for an unknown reason, despite the webpack versions being exactly the same. However, with a similar setup with slight differences in version, this would be reproducable.

If it's helpful to have a repro of this situation, I would be happy to spin up an issue with the repro and link it to this PR.

